PR #10, opened 2024-03-11
feat: add gh-actions to check and fix license headers

kengou:
Add a new GitHub Actions Workflow:
- Check & Fix License Header


.github/licenserc.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - 'LICENSE'
+    - 'NOTICE'
+    - 'PROJECT'
+    - '**/*.gitignore'
+    - '**/go.mod'
+    - '**/go.sum'
+    - '**/*.lock'
+    - '**/*.json'
+    - '**/.gitkeep'
+    - '**/*.jsx'
```
hodanoori, line 30: @kengou I just saw that you excluded .jsx files from being commented. What is the reason for that? We have a lot of files in our project which are written in .jsx files.

edda: Good catch @hodanoori! `.jsx` files should not be excluded

kengou: Fixing it in #27

PR #16, opened 2024-03-17
fix licenses; parameterize headscale chart

auhlig:


LICENSES/Apache-2.0.txt
```diff
@@ -58,7 +58,7 @@ APPENDIX: How to apply the Apache License to your work.
 
 To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information. (Don't include the brackets!)  The text should be enclosed in the appropriate comment syntax for the file format. We also recommend that a file or class name and description of purpose be included on the same "printed page" as the copyright notice for easier identification within third-party archives.
 
-Copyright [yyyy] [name of copyright owner]
+Copyright 2024-2026 SAP SE or an SAP affiliate company and Greenhouse contributors
```
hodanoori, line 61: Hi @auhlig, Just a qucik note: This line must not be changed as its just an explanation in this Appendix of how you can adjust license info in your project. Therefore, we already discuss this with @edda that we must not change License text file and instead we add those SPDX identifiers to mention this exact copyright info for each file. So please revert this line of change in each and every repository which you changed it.

hodanoori (commented):
@auhlig  I know this is already merged, but if you agree, please correct it in another PR.

PR #41, opened 2024-03-26
enforce clusterName configured on pluginconfig

auhlig:
A PluginConfig is mapped to an onboarded cluster within an organization.
This PR enforces the `$PluginConfig.spec.clusterName` is set except for some to-be-migrated ones and the greenhouse organization

### DIFF
--- a/pkg/admission/pluginconfig_webhook.go
+++ b/pkg/admission/pluginconfig_webhook.go
@@ -94,7 +94,11 @@ func ValidateCreatePluginConfig(ctx context.Context, c client.Client, obj runtim
 	return nil, nil
 }
 
-func ValidateUpdatePluginConfig(ctx context.Context, c client.Client, _, obj runtime.Object) (admission.Warnings, error) {
+func ValidateUpdatePluginConfig(ctx context.Context, c client.Client, old, obj runtime.Object) (admission.Warnings, error) {
+	oldPluginConfig, ok := obj.(*greenhousev1alpha1.PluginConfig)
+	if !ok {
+		return nil, nil
+	}
 	pluginConfig, ok := obj.(*greenhousev1alpha1.PluginConfig)
 	if !ok {
 		return nil, nil
@@ -111,6 +115,11 @@ func ValidateUpdatePluginConfig(ctx context.Context, c client.Client, _, obj run
 		return nil, err
 	}
 	if err := validatePluginConfigForCluster(ctx, pluginConfig, c); err != nil {
+		return nil, err
+	}
+	if err := validateImmutableField(oldPluginConfig.Spec.ClusterName, pluginConfig.Spec.ClusterName,
+		field.NewPath("spec", "clusterName"),
+	); err != nil {
 		return nil, err
 	}
 	return nil, nil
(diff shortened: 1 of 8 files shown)


pkg/admission/utils.go
```diff
@@ -91,3 +91,12 @@ func (c *customValidator) ValidateDelete(ctx context.Context, obj runtime.Object
 	}
 	return c.validateDelete(ctx, c.Client, obj)
 }
+
+func isStringSliceContains(theStringSlice []string, theString string) bool {
+	for _, s := range theStringSlice {
+		if s == theString {
+			return true
```
IvoGoman, line 98: This could be replaced with https://pkg.go.dev/slices@go1.21.0#Contains

pkg/admission/pluginconfig_webhook.go
```diff
@@ -178,15 +184,18 @@ func validatePluginConfigOptionValues(pluginConfig *greenhousev1alpha1.PluginCon
 	return apierrors.NewInvalid(pluginConfig.GroupVersionKind().GroupKind(), pluginConfig.Name, allErrs)
 }
 
-func validatePluginConfigClusterExists(ctx context.Context, pluginConfig *greenhousev1alpha1.PluginConfig, c client.Client) error {
-	// TODO: Enforce clusterName on Plugins with backend.
-	//  We allow PluginConfigs without cluster reference during migration.
-	//  Later a PluginConfig with a backend must have a clusterName configured. Frontend-only plugins are allowed without a clusterName.
```
IvoGoman, line 184: This will break for the heureka plugin, which is frontend-only.

pkg/admission/pluginconfig_webhook.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/cloudoperators/greenhouse/pkg/helm"
 )
 
+// pluginsAllowedInCentralCluster is a list of plugins that are allowed to be installed in the central cluster.
+// TODO: Make this configurable on plugin level (AdminPlugin discussion) instead of maintaining a list here.
+var pluginsAllowedInCentralCluster = []string{
+	"alerts", "doop", "service-proxy", "teams2slack",
```
IvoGoman, line 27: Should we add `service-proxy` and `kubeconfig-generator` to the list?

IvoGoman (changes requested):

pkg/admission/utils.go
auhlig: of course!

pkg/admission/pluginconfig_webhook.go
auhlig: done

pkg/admission/pluginconfig_webhook.go
auhlig: i excluded fontend-only plugins from the list

IvoGoman (approved):

auhlig:
I struggled a bit with the tests. They're fixed now and this can be reviewed and merged if deemed good. Thanks ❤️

IvoGoman (approved):

PR #47, opened 2024-03-27
Fix dependencies. Don't use latest

edda:
"Latest" doesn't work for dependencies. If they change then the sha hash of the version originally used and stored in `package-lock.json` doesn't match the sha hash of the current version anymore which leads to inconsistencies.


TilmanHaupt (approved):
looks good

PR #79, opened 2024-04-03
Plugin admin icon

TilmanHaupt:
Ressource Status Icon is now shown in the Details View. For a unified design language the icons are used like in cluster-admin.

### DIFF
--- a/ui/plugin-admin/src/components/PluginDetail.js
+++ b/ui/plugin-admin/src/components/PluginDetail.js
@@ -53,17 +53,10 @@ const PluginDetail = () => {
       onClose={onPanelClose}
       size="large"
       heading={
-        plugin?.disabled ? (
-          <Stack gap="2">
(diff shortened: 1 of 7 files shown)


ui/plugin-admin/src/components/PluginDetail.js
```diff
@@ -50,7 +52,19 @@ const PluginDetail = () => {
       opened={!!showDetailsFor}
       onClose={onPanelClose}
       size="large"
-      heading={plugin?.disabled ? plugin?.name + " DISABLED" : plugin?.name}
+      heading={
+        plugin?.disabled ? (
```
edda, line 56: Why the special handling for disabled plugins here? The `PluginConditionIcon` component takes care of this, doesn't it?

edda (changes requested):

edda (approved):

PR #96, opened 2024-04-05
[plugin-admin] update dependency utils

TilmanHaupt:
Set dependency utils to version 1.1.6


hodanoori (commented):
I think we said that for these updates in dependecies, we don't increase the version. So can leave this part

hodanoori (approved):
lgtm

PR #98, opened 2024-04-05
configurable namespace for a pluginconfig release

### LINKED ISSUE #50: Feat: Make PluginConfig namespace configurable
As noted down in the internal repository by @IvoGoman:

With the current implementation all Plugins will be installed into the Greenhouse managed namespace within the remote clusters.
In order to have a more granular setup and to avoid a "mono-namespace" it should be possible to provide a namespace with the PluginConfig. To ensure that no helm releases are left-behind this namespace should be immutable and defaulted to the Greenhouse managed namespace.

auhlig:
Closes #50.

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -61,7 +61,7 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
 	}
 	// A release does not exist. Install it.
 	if !isReleaseExists {
-		log.FromContext(ctx).Info("installing release for plugin", "namespace", plugin.Namespace, "name", plugin.Name)
+		log.FromContext(ctx).Info("installing release for plugin", "namespace", plugin.GetReleaseNamespace(), "name", plugin.Name)
 		_, err = installRelease(ctx, local, restClientGetter, pluginDefinition, plugin, false)
 		return err
 	}
@@ -80,7 +80,7 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
 	if releaseStatus, ok := isCanReleaseBeUpgraded(latestRelease); !ok {
 		return fmt.Errorf("cannot upgrade release %s/%s in status %s", latestRelease.Namespace, latestRelease.Name, releaseStatus.String())
 	}
-	log.FromContext(ctx).Info("upgrading release", "namespace", plugin.Namespace, "name", plugin.Name)
+	log.FromContext(ctx).Info("upgrading release", "namespace", plugin.GetReleaseNamespace(), "name", plugin.Name)
 
 	c, err := clientutil.NewK8sClientFromRestClientGetter(restClientGetter)
 	if err != nil {
@@ -95,7 +95,7 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
 
 // UninstallHelmRelease removes the Helm release for the given Plugin.
 func UninstallHelmRelease(ctx context.Context, restClientGetter genericclioptions.RESTClientGetter, plugin *greenhousev1alpha1.Plugin) (releaseNotFound bool, err error) {
-	cfg, err := newHelmAction(restClientGetter, plugin.Namespace)
+	cfg, err := newHelmAction(restClientGetter, plugin.GetReleaseNamespace())
 	if err != nil {
 		return false, err
 	}
@@ -177,7 +177,7 @@ func ResetHelmReleaseStatusToDeployed(ctx context.Context, restClientGetter gene
 		return err
 	}
 
-	cfg, err := newHelmAction(restClientGetter, plugin.Namespace)
+	cfg, err := newHelmAction(restClientGetter, plugin.GetReleaseNamespace())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/helm/helm.go
```diff
@@ -328,15 +328,15 @@ func upgradeRelease(ctx context.Context, local client.Client, restClientGetter g
...
 		return nil, err
 	}
 	installAction := action.NewInstall(cfg)
 	installAction.ReleaseName = pluginConfig.Name
-	installAction.Namespace = pluginConfig.Namespace
-	// Namespaces are only created by an Organization.
-	installAction.CreateNamespace = false
+	installAction.Namespace = pluginConfig.GetReleaseNamespace()
+	// Create the target namespace if it doesn't exist.
+	installAction.CreateNamespace = true
```
auhlig, line 339: Note the change here. 
The logic adheres to the Helm behavior and creates the namespace, but doesn't delete it. 
Only the Helm release is managed by Greenhouse to natively integrate into existing setups.

pkg/admission/pluginconfig_webhook.go
```diff
@@ -68,6 +68,11 @@ func DefaultPluginConfig(ctx context.Context, c client.Client, obj runtime.Objec
 		return err
 	}
 	pluginConfig.Spec.OptionValues = optionValues
+
+	// Default the ReleaseNamespace to the organization namespace if not set.
```
auhlig, line 72: For follow-up PRs: We might also want to consider providing defaults in the Plugin. E.g. kube-system is a common home for ingress-nginx, cert-manager namespace usually hosts for digicert-issuer, cert-manager, etc.

pkg/apis/greenhouse/v1alpha1/pluginconfig_types.go
```diff
@@ -26,6 +26,10 @@ type PluginConfigSpec struct {
 
 	// ClusterName is the name of the cluster the pluginConfig is deployed to. If not set, the pluginConfig is deployed to the greenhouse cluster.
 	ClusterName string `json:"clusterName,omitempty"`
+
+	// ReleaseNamespace is the namespace in the remote cluster to which the backend is deployed.
+	// Defaults to the organization namespace if not set.
```
IvoGoman, line 31: This accommodates the planned change to use a special name for the managed namespace

```suggestion
	// Defaults to the Greenhouse managed namespace if not set.
```

pkg/apis/greenhouse/v1alpha1/pluginconfig_types.go
```diff
@@ -137,6 +141,15 @@ type PluginConfig struct {
 	Status PluginConfigStatus `json:"status,omitempty"`
 }
 
+// GetReleaseNamespace returns the release namespace of the pluginconfig.
+func (p *PluginConfig) GetReleaseNamespace() string {
+	// TODO: Remove this method once all PluginConfigs have the ReleaseNamespace set.
```
IvoGoman, line 146: Also in consideration of the suggestion of adding a default namespace to the _PluginDefinitions_
```suggestion
	// TODO: Replace this method with a Defaulter in the webhook once all PluginConfigs have the ReleaseNamespace set.
```

pkg/admission/pluginconfig_webhook.go
```diff
@@ -68,6 +68,11 @@ func DefaultPluginConfig(ctx context.Context, c client.Client, obj runtime.Objec
...
+	if pluginConfig.Spec.ReleaseNamespace == "" {
+		pluginConfig.Spec.ReleaseNamespace = pluginConfig.GetNamespace()
+	}
```
IvoGoman, line 75: This seems to contradict the validation that the field is immutable. I believe this validation can only be done once all `Plugins` have the `.spec.releaseNamespace` set

pkg/admission/pluginconfig_webhook.go
```diff
@@ -123,6 +128,12 @@ func ValidateUpdatePluginConfig(ctx context.Context, c client.Client, old, obj r
 	); err != nil {
 		return nil, err
 	}
+	// Release migration is currently not supported. Thus, the ReleaseNamespace is immutable.
+	if err := validateImmutableField(oldPluginConfig.Spec.ReleaseNamespace, pluginConfig.Spec.ReleaseNamespace,
```
auhlig, line 132:
```suggestion
	if err := validateImmutableField(oldPluginConfig.GetReleaseNamespace(), pluginConfig.GetReleaseNamespace(),
```

auhlig: True. That should fix the [mentioned contradiction](https://github.com/cloudoperators/greenhouse/pull/98#discussion_r15550505069) during migration

pkg/admission/pluginconfig_webhook.go
auhlig: True. The [below](https://github.com/cloudoperators/greenhouse/pull/98#discussion_r1555595741) addresses that. Or we introduce the immutability after the migration of existing instances since we have to touch the code again anyways

auhlig:
That post-#112 rebase was slightly painful

pkg/admission/pluginconfig_webhook.go
IvoGoman: Either way fine with me. Risk of this being manually changed seems quite low.

auhlig:
rebased & force-pushed

auhlig:
Getting `the input device is not a TTY` [in the license check](https://github.com/cloudoperators/greenhouse/actions/runs/8629784423/job/23654686186?pr=98#step:4:202).

auhlig:
rebased again to incorporate [ci fix](https://github.com/cloudoperators/greenhouse/pull/128/files#)

auhlig:
> There are some mentions of plugin.Namespace in calls to clientutil.NewRestClientGetterFromRestConfig(..) and helm.ExportNewHelmAction(..)
Could you adapt these to use plugin.ReleaseNamespace as well. This is to avoid unwanted side-effects if in the future the ReleaseNamespace on the test resources is changed.

Sure. I'll double check
Some `plugin.Namespace` references are to remain though, e.g. when resolving option values from a secret next to the plugin we need the plugin namespace, not the release namespace in the remote cluster, etc.

Issue #83, opened 2024-04-03
[FEAT] - Plugin Bundle/ Cluster Preset Concept


IvoGoman:
### Priority

(Medium) I'm annoyed but I'll live

### Description

`As a _ClusterAdmin_, I would like to maintain a Plugin preset for a type of Cluster, so that I do not have to configure every Plugin manually for every Cluster that is on-boarded`

An Organization can have multiple Kubernetes Clusters that should be set-up in a unified way. To reduce the toil of doing the Plugin configuration manually, it would be great if the ClusterAdmin could configure a bundle/preset. 
This bundle/preset should contain a list of Plugins, default configuration etc. .
It should be possible to override/ complement these values with values specific to the cluster. 

Acceptance criteria:
- [ ] Propose a concept and with an ADR 
- [ ] Implement the Solution

PR #99, opened 2024-04-05
pluginbundle proposal

auhlig:
as discussion and 1st very pragmatic approach for #83. @IvoGoman  

this PR adds a PluginBundle CRD (naming?) to have organization and cluster admins define a set of plugins to be run on all onboarded clusters.


pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
```diff
@@ -0,0 +1,45 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PluginBundleSpec defines the desired state of PluginBundle
+type PluginBundleSpec struct {
+	// ClusterSelector is a label selector to select the clusters the plugin bundle should be deployed to.
+	ClusterSelector metav1.LabelSelector `json:"clusterSelector"`
+
+	// Plugins is a list of plugins that should be deployed.
+	Plugins []PluginConfigSpec `json:"plugins"`
```
auhlig, line 16: Re-use existing or have a new struct here?

auhlig: make additive only initially?

pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
```diff
@@ -0,0 +1,45 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PluginBundleSpec defines the desired state of PluginBundle
+type PluginBundleSpec struct {
+	// ClusterSelector is a label selector to select the clusters the plugin bundle should be deployed to.
+	ClusterSelector metav1.LabelSelector `json:"clusterSelector"`
```
auhlig, line 13: make additive only initially?

pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// PluginBundleStatus defines the observed state of PluginBundle
+type PluginBundleStatus struct{}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// PluginBundle is the Schema for the pluginbundles API
+type PluginBundle struct {
```
auhlig, line 26: naming?

pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
auhlig: On 2nd thought: This should only be a single spec 
```suggestion
	PluginConfigSpec PluginConfigSpec `json:"pluginConfigSpec"`
```

In other words, we'd have a `kube-monitoring` bundle that would ensure only this plugin in all selected clusters.

pkg/controllers/pluginconfig/pluginbundle_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, pluginConfigSpec := range pluginBundle.Spec.Plugins {
+		for _, cluster := range selectedClusters.Items {
+			var pluginConfigForCluster = new(greenhousev1alpha1.PluginConfig)
+			pluginConfigForCluster.Namespace = pluginBundle.GetNamespace()
+			pluginConfigForCluster.Name = makeNameForPluginConfigFromPluginBundle(&pluginConfigSpec, &cluster)
+			result, err := clientutil.CreateOrPatch(ctx, r.Client, pluginConfigForCluster, func() error {
+				pluginConfigForCluster.Spec = pluginConfigSpec
+				pluginConfigForCluster.Spec.ClusterName = cluster.GetName()
+				return nil
+			})
```
IvoGoman, line 77: In the case that an PluginAdmin changes the `OptionValues` for the generated `Plugin` it might not be obvious why the values are overwritten from time to time.
I can image three solutions with increasing effort:

- Only create _Plugins_ if they are non-existing
- Make it obvious to the enduser where they need to perform changes to ensure the changes are persistent
- Annotate the `Plugins` created by this controller. This way the annotation can be removed if the Plugin shall no longer be managed by this controller

pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
IvoGoman: A `PluginPreset` contains default _OptionValues_ for a _PluginDefinition_ that is automatically created for a _Cluster_, if this cluster is matched by the _ClusterSelector_, WDYT? 

The Bundle suggests more than one Plugin is contained.

auhlig: We can put the naming to the test in the next weekly

pkg/controllers/pluginconfig/pluginbundle_controller.go
IvoGoman: As discussed in the weekly meeting: Updates to Plugins that have been created by the PluginPreset controller will be forbidden. This will be enforced by the AdmissionWebhooks and the UI by checking for a `OwnerReference` on the _Plugin_ pointing to a _PluginPreset_

pkg/apis/greenhouse/v1alpha1/pluginbundle_types.go
IvoGoman: I went ahead with the renaming, as there were no strong opinions raised during the weekly.

pkg/controllers/plugin/pluginpreset_controller.go
```diff
@@ -0,0 +1,175 @@
...
+		// TODO: Handle the result. Log and emit event.
+		}
+		allErrs = append(allErrs, err)
+	}
+	return utilerrors.NewAggregate(allErrs)
+}
+
+// generatePluginName generates a name for a plugin based on the used PluginDefinition and the Cluster.
+func generatePluginName(pluginDefinition string, cluster *greenhousev1alpha1.Cluster) string {
+	return fmt.Sprintf("%s-%s", pluginDefinition, cluster.GetName())
```
auhlig, line 153: Only a remark on the naming here on whether to use `<pluginPreset.name>-<cluster.name>` to make it clearer what plugins belong to it.
Though, I see that `<pluginDefinition.name>-<cluster.name>` is also good to figure out what the actual plugin does.

What would happen, if a user already created a plugin with any of the naming conventions before creating the plugin preset?

pkg/controllers/pluginconfig/pluginbundle_controller.go
IvoGoman: Edits can't be forbidden, but will return a `Warning` and the PluginPresetController will reconcile the PluginPreset for which the dependent Plugin was updated.

pkg/controllers/plugin/pluginpreset_controller.go
IvoGoman: > Only a remark on the naming here on whether to use <pluginPreset.name>-<cluster.name> to make it clearer what plugins belong to it.
Though, I see that <pluginDefinition.name>-<cluster.name> is also good to figure out what the actual plugin does.

I like it. That makes it more obvious in the Plugin UI. The `kubectl get plugins` already contains both the _PluginDefinition_ and the _Cluster_ in the result.  

> What would happen, if a user already created a plugin with any of the naming conventions before creating the plugin preset?

Good point. I had not considered this yet. As a safeguard we could skip existing Plugins that do not have the `OwnerReference` and/or the `greenhouse.sap/pluginpreset` set.

kengou (changes requested):
Helm-lint action failed. Can you please take care of it ?

kengou (approved):

PR #118, opened 2024-04-09
Extend alerts plugin template by adding new attribute silenceTemplates.

ArtieReus:


charts/greenhouse/templates/pluginconfig-alerts.yaml
```diff
@@ -28,6 +28,9 @@ spec:
     - name: silenceExcludedLabels
       value:
         {{ .Values.alerts.silenceExcludedLabels | toYaml | nindent 8 }}
+    - name: silenceTemplates
+      value:
```
IvoGoman, line 32: We need to check if the `.Values.alerts.silenceTemplates` is set. In case it is not it will lead to Helm not being able to render this template.

PR #120, opened 2024-04-09
[cluster-admin] Highlighted list items for better UX

TilmanHaupt:
To ensure a similar user experience (UX), the following features are proposed:

- Minimum width column for the ResourceStatus
- Rows are now grey if the user hovers
- Rows are light grey if the detail view for the cluster is open
- If the user clicks the active list item the detail panel is closed.


edda (approved):
🙏

ui/cluster-admin/src/store.ts
```diff
@@ -35,8 +35,8 @@ export interface State {
   }
   setClusterDetails: (cluster: Cluster | null) => void
   setClusterDetailPlugins: (plugins: Plugin[]) => void
-  showClusterDetails: boolean
-  setShowClusterDetails: (showClusterDetails: boolean) => void
+  showClusterDetails: string | undefined
```
uwe-mayer, line 38: I would prefer to keep a boolean here for easier handling and reference.

ui/cluster-admin/src/components/ClusterListItem.tsx
```diff
@@ -46,7 +54,12 @@ const ClusterListItem: React.FC<ClusterListItemProps> = (
   }
 
   return (
-    <DataGridRow style={{ cursor: "pointer" }} onClick={() => openDetails()}>
+    <DataGridRow
+      className={`cursor-pointer ${
+        showClusterDetails === props?.cluster?.metadata?.name ? "active" : ""
```
uwe-mayer, line 59: The `clusterDetails.cluster` should already provide the necessary data. No need to use `showClusterDetails` here.

ui/cluster-admin/src/components/ClusterDetail.tsx
```diff
@@ -37,7 +37,7 @@ const ClusterDetail: React.FC<any> = () => {
   }, [clusterDetails.cluster?.metadata?.creationTimestamp])
 
   const onPanelClose = () => {
-    setShowClusterDetails(false)
+    setShowClusterDetails(undefined)
```
uwe-mayer, line 40: We need to also `null` the `cluster` prop in `clusterDetails` here to make sure the `active` state works correctly

uwe-mayer (changes requested):

ui/cluster-admin/src/components/ClusterDetail.tsx
uwe-mayer: So we need `setClusterDetails` from `useStore()` and set it here

uwe-mayer:
Thanks for adding this! Just some remarks on how to extract `active` state without changing the `showClusterDetails` type

TilmanHaupt:
> Thanks for adding this! Just some remarks on how to extract `active` state without changing the `showClusterDetails` type

Should be all fine now. :)

ui/cluster-admin/src/components/ClusterListItem.tsx
```diff
@@ -46,7 +55,15 @@ const ClusterListItem: React.FC<ClusterListItemProps> = (
   }
 
   return (
-    <DataGridRow style={{ cursor: "pointer" }} onClick={() => openDetails()}>
+    <DataGridRow
+      className={`cursor-pointer ${
+        clusterDetails?.cluster?.metadata?.name ===
+          props?.cluster?.metadata?.name && showClusterDetails
```
uwe-mayer, line 61: Nice, no need to `null` the `clusterDetails.cluster`

uwe-mayer (approved):

PR #126, opened 2024-04-10
Introduce git commit short sha as image tag

uwe-mayer:


.github/workflows/docker-build.yaml
```diff
@@ -98,6 +98,7 @@ jobs:
             type=ref,event=pr
             type=schedule
             type=raw,value=${{ github.sha }}
+            type=raw,value=${{ sha }}
```
kengou, line 101: I think what we need is: `type=sha,enable=true,format=short`

kengou (changes requested):

.github/workflows/docker-build.yaml
uwe-mayer, line 101:
```suggestion
            type=raw,value=sha
```

.github/workflows/docker-build.yaml
```diff
@@ -98,6 +98,7 @@ jobs:
             type=ref,event=pr
             type=schedule
             type=raw,value=${{ github.sha }}
+            type=raw,value=sha
```
uwe-mayer, line 101:
```suggestion
            type=sha,enable=true,format=short
```

kengou (approved):

PR #136, opened 2024-04-15
Fix/129/dockerbuild

kengou:


Dockerfile.dev-env
```diff
@@ -8,13 +12,13 @@ COPY . .
 # Build dev-env and setup-envtest
 RUN --mount=type=cache,target=/go/pkg/mod \
 	--mount=type=cache,target=/root/.cache/go-build \
-	make generate build-dev-env GO_ENABLED=0 GOOS=linux GOARCH=amd64 \
-	&& GOBIN=/workspace/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@bf15e44028f908c790721fc8fe67c7bf2d06a611 \
+	make generate build-dev-env CGO_ENABLED=${CGO_ENABLED} GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+	&& go install sigs.k8s.io/controller-runtime/tools/setup-envtest@bf15e44028f908c790721fc8fe67c7bf2d06a611 \
 	&& cp $(/workspace/bin/setup-envtest use ${ENVTEST_K8S_VERSION} -p path)/* /usr/local/bin
```
IvoGoman, line 17: `go install` will install to the `$GOPATH` directory by default. Could you adjust the copy to use the `$GOPATH` instead of the `/workspace/bin`

kengou: We have the `GOBIN` set in the `Makefile` as well: https://github.com/cloudoperators/greenhouse/blob/937069a71be716c5f126e7c72b90897f24f710b0/Makefile#L137-L140

kengou: reverted to original state with `GOBIN`

IvoGoman (approved):

PR #142, opened 2024-04-19
Feat/UI/introduce plugin overview

uwe-mayer:
This PR introduces the following updates to the `PluginDefinitions` type:

- `displayName`: A human-readable label for the pluginDefinition.
- `docMarkDownUrl`: An URL to the markdown documentation file for this `PluginDefinition`
-  `icon`: Icon specifies the icon to be used for this `PluginDefinition` in the Greenhouse UI. Icons can be either: 
    - A string representing a juno icon in camel case from [this list]( https://github.com/sapcc/juno/blob/main/libs/juno-ui-components/src/components/Icon/Icon.component.js#L6-L52)
    - A publicly accessable image reference to a .png file. Will be displayed 100x100px and grey scaled.

And some features to the `plugin-admin` view:

- A `PluginDefinitions` overview as `Available Plugins` Tab
- A `PluginDefinition` detail view 
- A `Plugin` edit/create view, currently accessible from 
  - the `Configure Plugin` button within a `PluginDefinition` detail view
  - clicking on a `Plugin` from the `Enabled Plugins` List within a `PluginDefinition` view

View [README](https://github.com/cloudoperators/greenhouse/blob/fceb8c8fa1a23cd2ca9cbe21ed26ae90f7771deb/ui/plugin-admin/README.md) on instructions on how to spin up the ui with local mock resources representing all introduced changes.

The former `plugindefinition-admin` has been merged into the `plugin-admin` the "least intrusive way". 

A couple of todos still need attention (not exclusive!):

- [ ] Merge two currently different stores into a single store
- [x] Actually create a `secret` from a `OptionValue` with type secret
- [x] Link the `edit` screen from the `Plugin` detail view
- [x] Open `Available Plugins` tab with respective `Plugin` detail after successful `Plugin` create/update
- [ ] Smooth scroll on `Panel` open within `PluginDefinition` screens
- [x] An option to delete a `Plugin`
- [ ] Use URLstate everywhere
- [ ] Hide empty entries in `Plugin` detail view. 
- [ ] Watch `Plugins` instead of [getting](https://github.com/cloudoperators/greenhouse/blob/main/ui/plugin-admin/src/hooks/useAPI.js#L89), so we get updates without browser reload
- [ ] Use a common interface for the `Plugin` object ([see here](https://github.com/cloudoperators/greenhouse/blob/main/ui/plugin-admin/src/hooks/useAPI.js#L55-L68))
- [x] Style imported markdown in the `PluginDefinition` detail view.
- [x] Check `TODO`s in code before merging
- [x] Set `docker-compose` image tags to `main` before merging.

Unchecked boxes are moved to post MVP here:
https://github.com/cloudoperators/greenhouse/issues/152


uwe-mayer:
cc @edda @TilmanHaupt

TilmanHaupt (approved):

.github/workflows/docker-build.yaml
```diff
@@ -114,7 +114,7 @@ jobs:
         #  cache-to: type=gha,mode=max
           platforms: |
             linux/amd64
-        #    linux/arm64
+            linux/arm64
```
kengou, line 117: If you uncomment it, the builds would fail. I'm on it to fix it for all of our images, but till that please keep it disabled

charts/manager/crds/greenhouse.sap_clusters.yaml
```diff
@@ -6,7 +6,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.12.1
```
kengou, line 9: In `main` we use `v0.14.0` controller-gen version. Can you please rebase ?

auhlig (approved):

.github/workflows/docker-build.yaml
uwe-mayer: Good catch!

charts/manager/crds/greenhouse.sap_clusters.yaml
uwe-mayer: good catch!

PR #191, opened 2024-04-30
feat(pluginpreset): make clusterSelector mutable

auhlig:


pkg/admission/pluginpreset_webhook.go
```diff
@@ -98,14 +98,6 @@ func ValidateUpdatePluginPreset(ctx context.Context, c client.Client, oldObj, cu
 		allErrs = append(allErrs, err)
 	}
 
-	if err := validateImmutableField(oldPluginPreset.Spec.ClusterSelector.String(), pluginPreset.Spec.ClusterSelector.String(), field.NewPath("spec", "clusterSelector")); err != nil {
-		allErrs = append(allErrs, err)
-	}
-
-	if err := validateImmutableField(oldPluginPreset.Spec.ClusterSelector.String(), pluginPreset.Spec.ClusterSelector.String(), field.NewPath("spec", "plugin", "clusterName")); err != nil {
```
IvoGoman, line 105: This part needs to stay, as we want to ensure that the clusterName is not set in the embedded spec of the plugin

IvoGoman (changes requested):
Thanks, just one small remark

IvoGoman (approved):

PR #233, opened 2024-05-23
Fix(ui): falsy string comparison + dot in option identifiers

uwe-mayer:
This fixes that check mark value changes were not reflected in Plugin updates on boolean values.

This was due to the fact that Javascript compares 
```
("" == false) == true
```

and we did not expect OptionValues to have dots in their identifiers as in e.g. `alerts.alertmanager.enabled`


uwe-mayer:
Superseeded by https://github.com/cloudoperators/greenhouse/pull/235

PR #257, opened 2024-06-04
fix(ui): only use watch to avoid rerenders in team-admin

uwe-mayer:


hodanoori (approved):
This section of code using client.get was primarily there as the watch does not cover detecting the incorrect URL and throw an error for that, which theoretically does not happen when user access team-admin via click on its icon in Greenhouse. The better option is to improve the watch function in client code to cover that case as well and then remove this part of the code.
But you can remove it because of rerendering issue which zou face now.

PR #258, opened 2024-06-04
Feat(manifests): Generate Typescript types from CRDs

uwe-mayer:
This adds a `make` target to generate the Typescript types ingested by the uis when recreating the `manifests` of the greenhouse crontroller.


IvoGoman (approved):

IvoGoman:
Noticed that there are changes to `schema.d.ts` when I run `make generate-types`. Should we add these changes to the PR?

uwe-mayer:
> Noticed that there are changes to `schema.d.ts` when I run `make generate-types`. Should we add these changes to the PR?

It's just reorder of the CRDs, right? Yes can always be committed, but make no difference imho.

ArtieReus (changes requested):
I would increase the version please :)

uwe-mayer:
> I would increase the version please :)

Version is updated :)

ArtieReus (approved):
lgtm

Issue #195, opened 2024-05-06
🐛 [BUG] - default for option type `secret` not working in PluginDefinition


uwe-mayer:
### Priority

(Low) Something is a little off

### Description

If there should be a `PluginDefinition` that provides a default value for an `option` of type `secret`, no `Plugin` is applicable with the default value.
Reason:
The webhook in charge of injecting default values will of course not inject
```
valueFrom
  secret:
    key: <some-key>
    name: <secret-name>
```
Nor will it create the necessary secret.

### Reproduction steps

```bash
1. Deploy blow `PluginDefinition` to your greenhouse cluster
2. Try deploying the belonging `Plugin` that makes use of the default value.
```

### Manifests

```yaml
apiVersion: greenhouse.sap/v1alpha1
kind: PluginDefinition
metadata:
  name: test-plugindefinition-with-secret-default
spec:
  displayName: Test Plugin with a defaulted secret otion
  description: Test Plugin with a defaulted secret otion
  version: "1.0.0"
  icon: description
  uiApplication:
    name: test-plugin-ui
    version: "latest"
  options:
    - name: defaultedSecretValue
      description: This value has a default
      required: false
      type: secret
      default: "myDefaultSecretValue"
yaml
apiVersion: greenhouse.sap/v1alpha1
kind: Plugin
metadata:
  name: test-plugin
  namespace: <your-namespace>
spec:
  pluginDefinition: test-plugindefinition-with-secret-default
  disabled: false
  clusterName: <some-cluster>
```
```


### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

PR #365, opened 2024-07-15
fix(PluginDefinitions): Disallow defaulting of Option with Type Secret

uwe-mayer:
## Description

This PR fixes https://github.com/cloudoperators/greenhouse/issues/195

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

### DIFF
--- a/pkg/admission/plugindefinition_webhook.go
+++ b/pkg/admission/plugindefinition_webhook.go
@@ -93,7 +93,7 @@ func validatePluginDefinitionOptionValueAndType(pluginDefinition *greenhousev1al
 		if err := option.IsValid(); err != nil {
 			return apierrors.NewInvalid(pluginDefinition.GroupVersionKind().GroupKind(), pluginDefinition.GetName(), field.ErrorList{
 				field.Invalid(field.NewPath("spec").Child("options").Child("name"), option.Name,
-					"A PluginOption Default must match the specified Type and no defaults are allowed in PluginOptions of Type Secret."),
+					"A PluginOption Default must match the specified Type, and defaults are not allowed in PluginOptions of the 'Secret' type."),
 			})
 		}
 	}
--- a/pkg/admission/plugindefinition_webhook.go
+++ b/pkg/admission/plugindefinition_webhook.go
@@ -93,7 +93,7 @@ func validatePluginDefinitionOptionValueAndType(pluginDefinition *greenhousev1al
 		if err := option.IsValid(); err != nil {
 			return apierrors.NewInvalid(pluginDefinition.GroupVersionKind().GroupKind(), pluginDefinition.GetName(), field.ErrorList{
 				field.Invalid(field.NewPath("spec").Child("options").Child("name"), option.Name,
-					"A PluginOption Default must match the specified Type."),
+					"A PluginOption Default must match the specified Type and no defaults are allowed in PluginOptions of Type Secret."),
 			})
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/apis/greenhouse/v1alpha1/plugindefinition_types.go
```diff
@@ -111,6 +111,15 @@ func (p *PluginOption) IsValid() error {
 	case PluginOptionTypeMap:
 		var m map[string]any
 		return json.Unmarshal(p.Default.Raw, &m)
+	case PluginOptionTypeSecret:
+		var s string
+		if err := json.Unmarshal(p.Default.Raw, &s); err != nil {
+			return err
+		}
+		if s != "" {
```
IvoGoman, line 119: It is sufficient to check for the length of `Default.Raw` to make sure the value was not set
```suggestion
		if len(p.Default.Raw) != 0 {
```

pkg/admission/plugindefinition_webhook.go
```diff
@@ -93,7 +93,7 @@ func validatePluginDefinitionOptionValueAndType(pluginDefinition *greenhousev1al
 		if err := option.IsValid(); err != nil {
 			return apierrors.NewInvalid(pluginDefinition.GroupVersionKind().GroupKind(), pluginDefinition.GetName(), field.ErrorList{
 				field.Invalid(field.NewPath("spec").Child("options").Child("name"), option.Name,
-					"A PluginOption Default must match the specified Type."),
+					"A PluginOption Default must match the specified Type and no defaults are allowed in PluginOptions of Type Secret."),
```
IvoGoman, line 96: NIT: Just a suggestion for the wording.
```suggestion
					"A PluginOption Default must match the specified Type, and defaults are not allowed in PluginOptions of the 'Secret' type."),
```

IvoGoman (changes requested):
One remark to checking if the Raw JSON is empty for the Secret type.

pkg/apis/greenhouse/v1alpha1/plugindefinition_types.go
uwe-mayer: :heavy_check_mark:

IvoGoman (approved):

pkg/apis/greenhouse/v1alpha1/plugindefinition_types.go
uwe-mayer: Hmm tests fail after this merge.

The `Raw` representation for 
- `nil` is `[]uint8 len: 4, cap: 8, [110,117,108,108]` (a.k.a. `"null"`)   
- `""` is  `[]uint8 len: 2, cap: 8, [34,34]`  (a.k.a. `"\"\""`)

Both `len(p.Default.Raw)>0` if we do not unmarshal the json first. I think this was the reason why I added the unmarshalling in the first place...

I added a second (also failing) test to illustrate. 
We could stop using `json.Marshal(val)` for `val==nil` in the tests, which is done by the `test.MustReturnJSONFor()` func.
But I would prefer to catch the wider range of _not defaulted_ string values?

IvoGoman: My Bad, I had just tested this with by converting `""` to `[]byte` and that looked good. Yeah let's do it the way you have proposed intiially, then we are on the safe side.

PR #372, opened 2024-07-17
fix(teamRBAC): ensure status updated on cluster not found

IvoGoman:
## Description

If the cluster cannot be found by name the ClusterListEmpty status should be updated. Previous the reconcile failed without updating the status. Ready condition stayed true

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # (issue)
- Closes # (issue)
- Fixes # (issue)

> Remove if not applicable

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist

### DIFF
--- a/pkg/controllers/teamrbac/teamrolebinding_controller.go
+++ b/pkg/controllers/teamrbac/teamrolebinding_controller.go
@@ -102,7 +102,7 @@ func (r *TeamRoleBindingReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	defer func() {
 		if statusErr := r.setStatus(ctx, trb, trbStatus); statusErr != nil {
-			log.FromContext(ctx).Error(statusErr, "Error setting status for TeamRoleBinding", "TeamRoleBinding", trb.GetName())
+			log.FromContext(ctx).Error(statusErr, "Error setting status for TeamRoleBinding")
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -122,6 +122,7 @@ func (r *TeamRoleBindingReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	// list the clusters that either match the ClusterName or the ClusterSelector
 	clusters, err := r.listClusters(ctx, trb)
 	if err != nil {
+		trbStatus.SetConditions(greenhousev1alpha1.TrueCondition(greenhousev1alpha1.ReconcileFailedReason, "Failed to get cluster for TeamRoleBinding", ""))
```
uwe-mayer, line 125: You are passing the `greenhousev1alpha1.ReconcileFailedReason` as `greenhousev1alpha1.ConditionType` here. 
If I am not mistaken this is only possible and not caught because of missing types in all reasons here:
https://github.com/cloudoperators/greenhouse/blob/main/pkg/apis/greenhouse/v1alpha1/teamrolebinding_types.go#L99-L106

uwe-mayer (changes requested):

pkg/controllers/teamrbac/teamrolebinding_controller.go
IvoGoman: This was deliberate. While implementing the TeamRoleBindings I had noticed that we were creating lots of `<CRD>ReconcileFailed` variables. Both, the `events` and the `controller logs` already contain the context of the CRD. So I think we should go towards more generic reason/conditions instead of duplicating them for each CRD.

In this case that the error is not explicit I used the very generic `ReconcileFailedReason`. Which is strictly from the naming of the constant counterintuitive.

uwe-mayer: Not sure I completely understand...
So `ReconcileFailedReason` is a [ConditionReason](https://pkg.go.dev/github.com/cloudoperators/greenhouse/pkg/apis/greenhouse/v1alpha1#ConditionReason) or is it a [ConditionType](https://pkg.go.dev/github.com/cloudoperators/greenhouse/pkg/apis/greenhouse/v1alpha1#ConditionType)?

Looking at the signature of [`greenhousev1alpha1.TrueCondition()`](https://github.com/cloudoperators/greenhouse/blob/main/pkg/apis/greenhouse/v1alpha1/conditions.go#L64) it is used as `ConditionType` here.

I think it should not be usable as both, don't you think?
So we should improve typing in https://github.com/cloudoperators/greenhouse/blob/main/pkg/apis/greenhouse/v1alpha1/teamrolebinding_types.go#L99-L106. 

Compare e.g. to https://github.com/cloudoperators/greenhouse/blob/main/pkg/apis/greenhouse/v1alpha1/teamrolebinding_types.go#L79-L96

I am approving, so I don't block while I am on vacation...

uwe-mayer (approved):

pkg/controllers/teamrbac/teamrolebinding_controller.go
IvoGoman: I see now... That should have been the different type. The `ReconcileFailedReason` should just be a `string`, as it was used for the events. Where there is no strict typing for the `eventtype`.

The latest commit moves the generic event types into a own _events_types.go_. 
Also events for deleting resources are only emitted overall to avoid potential spamming of events. Instead CRUD actions on individual resources created by the controller are logged.
Furthermore, the logging is improved to remove fields already logged from the context.

PR #495, opened 2024-08-27
feat(pluginpreset) Skip updating when plugin is not different (#437)

### LINKED ISSUE #437: [FEAT] - PluginPreset only update Plugin's if the spec has changed
### Priority

Medium

### Description

The PluginPreset controller will patch all Plugins for a PluginPreset even if only one has been changed.
To reduce the unnecessary reconciliations the PluginPreset controller should check if there are changed to the Plugin's spec before it calls `clientutil.CreateOrPatch(..)`.  This way only those plugins are updated where the `.spec` no longer matches what the `PluginPreset` expects.

## Acceptance Criteria:

- [ ] Add a comparison between `Plugin.spec` and `PluginPreset.spec.plugin`, ignoring Defaults from the `PluginDefinition` and special `global.greenhouse` values
- [ ] Only update a Plugin if there are changes
- [ ] Test the feature 

### Reference Issues

_No response_

gciezkowski-acc:
## Description
Add skipping of updating if plugin preset has not been changed.

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # (issue)
- Closes #437
- Fixes # (issue)

> Remove if not applicable

## Added tests?

- [X] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [X] 🙅 no documentation needed

## Checklist

- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pkg/controllers/plugin/pluginpreset_controller.go
+++ b/pkg/controllers/plugin/pluginpreset_controller.go
@@ -45,6 +45,7 @@ type PluginPresetReconciler struct {
 //+kubebuilder:rbac:groups=greenhouse.sap,resources=pluginpresets/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=greenhouse.sap,resources=pluginpresets/finalizers,verbs=update
 //+kubebuilder:rbac:groups=greenhouse.sap,resources=plugins,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=greenhouse.sap,resources=plugindefinitions,verbs=get
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *PluginPresetReconciler) SetupWithManager(name string, mgr ctrl.Manager) error {
@@ -159,6 +160,13 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
 	failedCondition = *status.GetConditionByType(greenhousev1alpha1.PluginFailedCondition)
 	skippedCondition = *status.GetConditionByType(greenhousev1alpha1.PluginSkippedCondition)
 
+	pluginDefinition := &greenhousev1alpha1.PluginDefinition{}
+	err = r.Get(ctx, client.ObjectKey{Name: preset.Spec.Plugin.PluginDefinition}, pluginDefinition)
+	if err != nil {
+		allErrs = append(allErrs, err)
+		return skippedCondition, failedCondition, utilerrors.NewAggregate(allErrs)
+	}
+
 	for _, cluster := range clusters.Items {
 		plugin := &greenhousev1alpha1.Plugin{}
 		err := r.Get(ctx, client.ObjectKey{Namespace: preset.GetNamespace(), Name: generatePluginName(preset, &cluster)}, plugin)
@@ -166,7 +174,7 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
 		switch {
 		case err == nil:
 			// The Plugin exists but does not contain the labels of the PluginPreset. This Plugin is not managed by the PluginPreset and must not be touched.
-			if shouldSkipPlugin(plugin, preset) {
+			if shouldSkipPlugin(plugin, preset, pluginDefinition) {
 				skippedPlugins = append(skippedPlugins, plugin.Name)
 				continue
 			}
@@ -221,7 +229,7 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
 	return skippedCondition, failedCondition, utilerrors.NewAggregate(allErrs)
 }
 
-func shouldSkipPlugin(plugin *greenhousev1alpha1.Plugin, preset *greenhousev1alpha1.PluginPreset) bool {
+func shouldSkipPlugin(plugin *greenhousev1alpha1.Plugin, preset *greenhousev1alpha1.PluginPreset, definition *greenhousev1alpha1.PluginDefinition) bool {
 	if plugin.Labels[greenhouseapis.LabelKeyPluginPreset] != preset.Name {
 		return true
 	}
@@ -234,7 +242,21 @@ func shouldSkipPlugin(plugin *greenhousev1alpha1.Plugin, preset *greenhousev1alp
 		}
 	}
 
-	return len(preset.Spec.Plugin.OptionValues) >= countWithoutGlobalOption(plugin.Spec.OptionValues)
+	definitionCount := 0
+	for _, definitionOption := range definition.Spec.Options {
+		if definitionOption.Default == nil {
+			continue
+		}
+
+		definitionCount++
+		if !slices.ContainsFunc(plugin.Spec.OptionValues, func(item greenhousev1alpha1.PluginOptionValue) bool {
+			return item.Name == definitionOption.Name && string(item.Value.Raw) == string(definitionOption.Default.Raw)
+		}) {
+			return false
+		}
+	}
+
+	return len(preset.Spec.Plugin.OptionValues)+definitionCount >= countWithoutGlobalOption(plugin.Spec.OptionValues)
 }
 
 func countWithoutGlobalOption(list []greenhousev1alpha1.PluginOptionValue) int {
@@ -243,7 +265,6 @@ func countWithoutGlobalOption(list []greenhousev1alpha1.PluginOptionValue) int {
 		if strings.HasPrefix(item.Name, "global.greenhouse") {
 			continue
 		}
-
 		count++
 	}
 
--- a/pkg/controllers/plugin/pluginpreset_controller.go
+++ b/pkg/controllers/plugin/pluginpreset_controller.go
@@ -234,6 +234,7 @@ func shouldSkipPlugin(plugin *greenhousev1alpha1.Plugin, preset *greenhousev1alp
 		return true
 	}
 
+	// need to reconcile when plugin does not have option which exists in plugin preset
 	for _, presetOptionValue := range preset.Spec.Plugin.OptionValues {
 		if !slices.ContainsFunc(plugin.Spec.OptionValues, func(item greenhousev1alpha1.PluginOptionValue) bool {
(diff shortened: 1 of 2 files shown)


pkg/controllers/plugin/pluginpreset_controller.go
```diff
@@ -218,6 +218,47 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
 	return skippedCondition, failedCondition, utilerrors.NewAggregate(allErrs)
 }
 
+func shouldSkipPlugin(plugin *greenhousev1alpha1.Plugin, preset *greenhousev1alpha1.PluginPreset) bool {
+	if plugin.Labels[greenhouseapis.LabelKeyPluginPreset] != preset.Name {
+		return true
+	}
+
+	if preset.Spec.Plugin.PluginDefinition != plugin.Spec.PluginDefinition {
```
uwe-mayer, line 226: This is a valid thought, but we should deny changes in `Plugin.Spec.PluginDefinition` after Plugin `CREATE` in a webhook, wdyt @IvoGoman ?

pkg/controllers/plugin/pluginpreset_controller.go
```diff
@@ -218,6 +218,47 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
...
+		return false
+	}
+
+	for _, presetOptionValue := range preset.Spec.Plugin.OptionValues {
+		if !containsPluginOption(plugin.Spec.OptionValues, presetOptionValue) {
+			return false
+		}
+	}
+
+	return len(preset.Spec.Plugin.OptionValues) >= countWithoutGlobalOption(plugin.Spec.OptionValues)
```
uwe-mayer, line 236: This would not take `PluginDefinition.Options` into account that transport a `default` value to the `Plugin.OptionValues`, right? We are only excluding by prefix `global.greenhouse`...
Can you add a test case for a `PluginDefinition` with a defaulted `Option`?

pkg/controllers/plugin/pluginpreset_controller_test.go
```diff
@@ -242,6 +242,260 @@ var _ = Describe("PluginPreset Controller Lifecycle", Ordered, func() {
 	})
 })
 
+var _ = Describe("Plugin Preset skip changes", Ordered, func() {
```
uwe-mayer, line 245: This feels like a table test...
Have a look [here](https://github.com/cloudoperators/greenhouse/blob/main/pkg/controllers/propagation_reconciler_test.go#L141-L244)

uwe-mayer: And maybe we can make use of and extend the [test utilities / resources](https://github.com/cloudoperators/greenhouse/blob/main/pkg/test/resources.go)

Find [some examples in these tests](https://github.com/cloudoperators/greenhouse/blob/main/pkg/controllers/teamrbac/teamrolebinding_controller_test.go)

pkg/controllers/plugin/pluginpreset_controller_test.go
```diff
@@ -242,6 +242,260 @@ var _ = Describe("PluginPreset Controller Lifecycle", Ordered, func() {
...
+						},
+					},
+				},
+			},
+		}
+
+		Expect(shouldSkipPlugin(testPlugin, testPresetCase)).To(BeFalse())
+	})
+
+	It("should skip when plugin preset has the same values like plugin", func() {
```
uwe-mayer, line 456: Let's add a test for defaulted values as mentioned here... :)

uwe-mayer (dismissed):
:rocket: 

Some comments :)

pkg/controllers/plugin/pluginpreset_controller.go
IvoGoman: I think we should only care on the PluginPreset side that the values part of the PluginPreset Spec are also set on the Plugin. We can ignore everything else imo. The other values are set by the Webhooks and are not a concern for the PresetController

pkg/controllers/plugin/pluginpreset_controller.go
IvoGoman: This validation is already in Place, so no need to check this here: https://github.com/cloudoperators/greenhouse/blob/ecba94afc093967434d260b75123a6bb7174bf59/pkg/admission/pluginpreset_webhook.go#L97-L99
But it seems this is not the case for the Plugin itself 👀 https://github.com/cloudoperators/greenhouse/blob/main/pkg/admission/plugin_webhook.go#L105

IvoGoman: I am taking care of updating the webhook, as there is some more cleanup to be done as well.

pkg/controllers/plugin/pluginpreset_controller.go
uwe-mayer: Meaning we can loose the `len()` check, right?

pkg/controllers/plugin/pluginpreset_controller.go
gciezkowski-acc: Done

pkg/controllers/plugin/pluginpreset_controller_test.go
gciezkowski-acc: Done

pkg/controllers/plugin/pluginpreset_controller.go
gciezkowski-acc: Without that check the test fails [here](https://github.com/cloudoperators/greenhouse/blob/2c1fe55b5214a20b3b80f2f0af1193e408ce63b5/pkg/controllers/plugin/pluginpreset_controller_test.go#L131).

pkg/controllers/plugin/pluginpreset_controller.go
```diff
@@ -218,6 +218,47 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
...
+	if plugin.Labels[greenhouseapis.LabelKeyPluginPreset] != preset.Name {
+		return true
+	}
+
+	if preset.Spec.Plugin.PluginDefinition != plugin.Spec.PluginDefinition {
+		return false
+	}
+
+	for _, presetOptionValue := range preset.Spec.Plugin.OptionValues {
+		if !containsPluginOption(plugin.Spec.OptionValues, presetOptionValue) {
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/plugin/pluginpreset_controller_test.go
gciezkowski-acc: I added the test for defaulted values in PluginPreset lifecycle test.

pkg/controllers/plugin/pluginpreset_controller.go
IvoGoman: The assumption that everything besides the PluginPreset spec can be ignored was wrong. I had not considered the case where the managed Plugin was changed.
The failing testcase checks this condition that a PluginOptionValue added to the managed Plugin is reverted by the PluginPreset controller.

So it should not skip the Plugin if there are PluginOptionValues that are neither declared on the PluginPreset nor on the PluginDefinition.

gciezkowski-acc: Done

IvoGoman (changes requested):
Sorry for taking this into another round. Could you double check, but I think there are two errors in the test cases. They bring out the same expected result, but are not the how the test cases are described.

The loop is just for your consideration. I was thinking how/if we could not rely on counting the options.

pkg/controllers/plugin/pluginpreset_controller_test.go
```diff
@@ -242,6 +275,380 @@ var _ = Describe("PluginPreset Controller Lifecycle", Ordered, func() {
...
+						greenhouseapis.LabelKeyPluginPreset: pluginPresetName,
+					},
+				},
+				Spec: greenhousev1alpha1.PluginSpec{
+					PluginDefinition: pluginPresetDefinitionName,
+					OptionValues: []greenhousev1alpha1.PluginOptionValue{
+						{
+							Name:  greenhouseapis.GroupName + ".test_parameter",
+							Value: asAPIextensionJSON(2),
+						},
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/plugin/pluginpreset_controller_test.go
```diff
@@ -242,6 +275,380 @@ var _ = Describe("PluginPreset Controller Lifecycle", Ordered, func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						greenhouseapis.LabelKeyPluginPreset: pluginPresetName,
+					},
+				},
+				Spec: greenhousev1alpha1.PluginSpec{
+					PluginDefinition: pluginPresetDefinitionName,
+					OptionValues: []greenhousev1alpha1.PluginOptionValue{
+						{
+							Name:  "greenhouse.global.test_parameter",
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/plugin/pluginpreset_controller.go
```diff
@@ -218,6 +226,48 @@ func (r *PluginPresetReconciler) reconcilePluginPreset(ctx context.Context, pres
...
+	count := 0
+	for _, item := range list {
+		if strings.HasPrefix(item.Name, "global.greenhouse") {
+			continue
+		}
+		count++
+	}
+
+	return count
+}
```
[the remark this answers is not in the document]
gciezkowski-acc: I restructured the code into two loops. Your approach does not cover the situation where we have an option in the plugin preset that does not exist in the plugin.
Done.

IvoGoman (approved):

PR #548, opened 2024-09-06
fix(ui): propagate assetsHost to appLoader

andypf:
## Description

This PR fixes a bug in org-admin. During the removal of the widget loader from both the Dashboard and org-admin, a new library was created in both packages to load a module either from the assets server or the local server. Unlike the previous widget loader, which always loaded the import map from the assets server first, this new library does not load an import map. Instead, we now use a static import map for essential libraries like React, while remote modules are dynamically loaded using vanilla JavaScript. This simplifies the loading logic and enables us to offer all-in-one builds.

During this transition, a bug was introduced in the new library: the assetsHost property was not being used, which caused the system to always attempt to load the modules from the current server.

The issue was resolved by properly propagating the assetsHost property.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Checklist


ui/org-admin/secretProps.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+module.exports = {
```
edda, line 5: Looks like this file is missing in the `.gitignore`

edda (changes requested):

edda (approved):

PR #616, opened 2024-10-01
fix(types): Reintroduce typescript types

uwe-mayer:
## Description
This reintroduces the `ui` folder with the generated Typescript `schema.d.ts` used for typescript creation on CRD change.


uwe-mayer:
> wouldn't make sense to host it into a different folder then UI now? Just to keep the context clear i was thinking.

True. Even more if we might export typing for other clients. Moved to `./types/typescript`

abhijith-darshan (approved):

IvoGoman (approved):

PR #617, opened 2024-10-01
fix(service-proxy): Correctly decompose host

uwe-mayer:
## Description
We introduced a regression [with changing the order of parameters](https://github.com/cloudoperators/greenhouse/commit/9a049d83879e41e6e0009f9fa7f910ca33c31308#diff-d3666ab886c1a27806b6d26677a42b167e6816af8c07fcade3bb8328f115f7dc) in the service-proxy host.

This PR hotfixes this regression, but we still need a better permanent solution

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

### DIFF
--- a/pkg/common/url.go
+++ b/pkg/common/url.go
@@ -6,6 +6,7 @@ package common
 import (
 	"crypto/sha256"
 	"fmt"
+	"strings"
 
 	greenhousev1alpha1 "github.com/cloudoperators/greenhouse/pkg/apis/greenhouse/v1alpha1"
 )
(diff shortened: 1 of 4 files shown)


pkg/common/url.go
```diff
@@ -27,3 +28,19 @@ func URLForExposedServiceInPlugin(serviceName string, plugin *greenhousev1alpha1
...
+	parts := strings.SplitN(host, ".", 2)
+	if len(parts) < 2 {
+		return "", "", "", fmt.Errorf("invalid host: %s", host)
+	}
+	parts = strings.SplitN(parts[0], "--", 3)
+	if len(parts) < 3 {
+		return "", "", "", fmt.Errorf("invalid host: %s", host)
+	}
+	return parts[0], parts[1], parts[2], nil
+}
```
abhijith-darshan, line 46: Just a Suggestion - 

Would you prefer to return a struct instead and a verbose error on where it failed which can logged? This way we may not have to deal with multiple string and order?

Example:

```go
type ServiceProxyHost struct {
	Name      string
	Cluster   string
	Namespace string
}
```

```go
func SplitHost(host string) (ServiceProxyHost, error) {
	// Split the host into the name part and domain part
	parts := strings.SplitN(host, ".", 2)
	if len(parts) < 2 {
		return ServiceHost{}, fmt.Errorf("invalid host: missing domain: %s", host)
	}

	// Further split the first part (before the first dot) into service, cluster, and namespace
	parts = strings.SplitN(parts[0], "--", 3)
	if len(parts) < 3 {
		return ServiceHost{}, fmt.Errorf("invalid host: expected format $service--$cluster--$namespace: %s", host)
	}

	return ServiceProxyHost{
		Name:      nameParts[0],
		Cluster:   nameParts[1],
		Namespace: nameParts[2],
	}, nil
}
```

abhijith-darshan (approved):

pkg/common/url.go
uwe-mayer: Yes that makes a lot of sense. I would like to create an issue to properly tackle a refactoring.
This now just moves the relevant functions into one file for easier reference and hot fixes the faulty order.

PR #634, opened 2024-10-10
feat(cluster) configurable max token validity (#518)

### LINKED ISSUE #518: [FEAT] - Configurable maxTokenValidity for Cluster
### Priority

(Low) Something is a little off

### Description

> As a ClusterAdmin I would like to configure the maximal token validity for the Greenhouse ServiceUser KubeConfig, so that I can hibernate my cluster over the weekend without having to manually refresh the token.

Gardener clusters can be set into hibernation to save costs when they are not used. Currently all clusters have a maximal token validity of 24h. When a cluster is hibernating longer than 24h the token will expire and Greenhouse will no longer be able to access the cluster.
As a solution to this problem the `maxTokenValidity` must become configurable.

## Acceptance Criteria

- [ ] Cluster CRD receives an additional field `kubeConfig.maxTokenValidity`
- [ ] Cluster webhook to default this new value
- [ ] Maximum allowed value is 72h (for now)
- [ ] ❗ this will break existing clusters, as defaulting for existing CRs only takes place after changing them. Which requires manual intervention after deployment of the changes. Best case would be to introduce a new `v1alpha2` version for the Cluster CRD and add Conversions. Otherwise, the controller must be updated to temporarily handle the missing default value.

### Reference Issues

_No response_

gciezkowski-acc:
## Description
I added the configurable parameter max token validity to the cluster specification, ensuring it does not negatively impact existing clusters.

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # (issue)
- Closes #518
- Fixes # (issue)

> Remove if not applicable

## Added tests?

- [X] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist

- [x] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works


pkg/admission/cluster_webhook.go
```diff
@@ -47,6 +47,8 @@ func DefaultCluster(ctx context.Context, _ client.Client, obj runtime.Object) er
 	if !ok {
 		return nil
 	}
+	cluster.SetDefaultTokenValidityIfNeeded()
```
IvoGoman, line 50: The `kubebuilder:default` annotation defines a default value for the `MaxTokenValidity`. Do we need to default it again? Could you validate with a test if the defaulting works with just the annotation set?

pkg/apis/greenhouse/v1alpha1/cluster_types.go
```diff
@@ -4,13 +4,19 @@
...
 )
 
 // ClusterSpec defines the desired state of the Cluster.
 type ClusterSpec struct {
 	// AccessMode configures how the cluster is accessed from the Greenhouse operator.
 	AccessMode ClusterAccessMode `json:"accessMode"`
+
+	// MaxTokenValidity specifies the maximum duration for which a token remains valid
+	// +kubebuilder:default:="72h"
+	MaxTokenValidity metav1.Duration `json:"maxTokenValidity,omitempty"`
```
IvoGoman, line 19: I think we should add some structure here by using a custom struct field`KubeConfig` to contain the `MaxTokenValidity`. This way we could also add other KubeConfig specific configuration such as Token secret reference in the feature. 
@kengou what do you think?

kengou: That is a good idea

pkg/admission/cluster_webhook.go
```diff
@@ -160,3 +163,12 @@ func ValidateDeleteCluster(ctx context.Context, _ client.Client, obj runtime.Obj
 	logger.Error(err, "deletion request denied", "cluster", cluster.GetName())
 	return admission.Warnings{"deletion is not allowed"}, err
 }
+
+func validateTokenValidity(cluster *greenhousev1alpha1.Cluster) (admission.Warnings, error) {
+	if cluster.Spec.MaxTokenValidity.Duration > greenhousev1alpha1.MaxTokenValidity {
+		err := apierrors.NewBadRequest("token validity is too long")
+		return admission.Warnings{"token validity too long"}, err
+	}
```
kengou, line 171: I would probably overwrite the `MaxTokenValidity` to the default `72 hours` if the value is bigger, just to not block the user to onboard the cluster. Also somehow present the warning and explain that we did overwrite his value ( either UI and or via `greenhousectl`
@IvoGoman wdyt ?

kengou (changes requested):

pkg/admission/cluster_webhook.go
abhijith-darshan: We don't need to defaulting webhook if we use the `kubebuilder` default annotation

You can see this in the CRD manifest that is generated -

https://github.com/cloudoperators/greenhouse/pull/634/files#diff-51013bf768fe937b255fe3c3bb649b25a64efd134fc1293c59dfd3eb7d6a0046R1097-R1100

gciezkowski-acc: I have doubts so I will check it.

abhijith-darshan: ok 😸

pkg/admission/cluster_webhook.go
abhijith-darshan: Since validation is needed here (I forgot before suggesting default) then we need to make `MaxTokenValidity`as an `int32` instead and then we can do the following annotations on top of it -

```go
// ClusterSpec defines the desired state of the Cluster.
type ClusterSpec struct {
	// AccessMode configures how the cluster is accessed from the Greenhouse operator.
	AccessMode ClusterAccessMode `json:"accessMode"`

	// MaxTokenValidity specifies the maximum duration for which a token remains valid
	// +kubebuilder:default:=72
	// +kubebuilder:validation:Minimum=24
	// +kubebuilder:validation:Maximum=72
	MaxTokenValidity *int32 `json:"maxTokenValidity,omitempty"`
        ...
}
```

pkg/admission/cluster_webhook.go
gciezkowski-acc: I've checked it. The `kubebuilder:default` is not enough so we should have it here.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -111,10 +111,11 @@ func (r *RemoteClusterReconciler) EnsureCreated(ctx context.Context, resource li
 		return ctrl.Result{}, lifecycle.Failed, err
 	}
 
+	cluster.SetDefaultTokenValidityIfNeeded()
 	var tokenRequest = &tokenHelper{
 		Client:                             r.Client,
-		RemoteClusterBearerTokenValidity:   r.RemoteClusterBearerTokenValidity,
-		RenewRemoteClusterBearerTokenAfter: r.RenewRemoteClusterBearerTokenAfter,
+		RemoteClusterBearerTokenValidity:   cluster.Spec.KubeConfig.MaxTokenValidity.Duration,
+		RenewRemoteClusterBearerTokenAfter: cluster.Spec.KubeConfig.MaxTokenValidity.Duration - (24 * time.Hour),
```
IvoGoman, line 118: This could break because we have no minimum value for the MaxTokenValidity.
I would propose 24h minimum & default to be 72h.
Then we can keep the `r.RenewRemoteClusterBearerTokenAfter`

pkg/apis/greenhouse/v1alpha1/cluster_types.go
gciezkowski-acc: Done

pkg/admission/cluster_webhook.go
gciezkowski-acc: Done

pkg/controllers/cluster/cluster_controller.go
gciezkowski-acc: Done, I set minimum value.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -111,10 +111,11 @@ func (r *RemoteClusterReconciler) EnsureCreated(ctx context.Context, resource li
 		return ctrl.Result{}, lifecycle.Failed, err
 	}
 
+	cluster.SetDefaultTokenValidityIfNeeded()
 	var tokenRequest = &tokenHelper{
 		Client:                             r.Client,
-		RemoteClusterBearerTokenValidity:   r.RemoteClusterBearerTokenValidity,
-		RenewRemoteClusterBearerTokenAfter: r.RenewRemoteClusterBearerTokenAfter,
+		RemoteClusterBearerTokenValidity:   time.Duration(cluster.Spec.KubeConfig.MaxTokenValidity) * time.Hour,
+		RenewRemoteClusterBearerTokenAfter: (time.Duration(cluster.Spec.KubeConfig.MaxTokenValidity) * time.Hour) - (24 * time.Hour),
```
IvoGoman, line 118: In case `MaxTokenValidity` is set to 24h, this will lead to the token never being refreshed on every reconciliation.
Let's leave it with the old value of `r.RenewRemoteClusterBearerTokenAfter` so that it is renewed once 20h are left.

gciezkowski-acc: Done

pkg/admission/cluster_webhook.go
gciezkowski-acc: Done

IvoGoman (approved):

kengou (approved):

PR #653, opened 2024-10-23
Add alert for failing Plugin Helm Chart tests

ibakshay:
This pull request adds an alert for failing Helm Chart tests. The alert is triggered when the Helm Chart test has been failing for the last 15 minutes. The alert includes information about the plugin, namespace, and cluster of the remote cluster.

pre-requisite for https://github.com/cloudoperators/greenhouse-extensions/issues/447

### DIFF
--- a/charts/manager/alerts/operator.alerts
+++ b/charts/manager/alerts/operator.alerts
@@ -28,6 +28,15 @@ groups:
       for: 15m
       labels:
         severity: warning
+    - alert: GreenhousePluginHelmChartTestFailures
+      expr: |
+        sum by(plugin, cluster, namespace)(rate(greenhouse_plugin_chart_test_runs_total{result="Error"}[30m])) > 0
+      for: 15m
+      labels:
+        severity: warning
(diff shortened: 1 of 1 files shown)


charts/manager/alerts/operator.alerts
```diff
@@ -28,6 +28,15 @@ groups:
       for: 15m
       labels:
         severity: warning
+    - alert: GreenhousePluginHelmChartTestFailures
+      expr: |
+        sum by(plugin, cluster, namespace)(rate(greenhouse_plugin_chart_test_runs_total{result="Error"}[30m])) > 0
```
IvoGoman, line 33: The range of the rate should be less than the `for` condition. This would fire as long as one test in the last 30m failed.

ibakshay: thanks a lot! fixed in 3869b3f3d7de8beef50c6c10575a12d1d5272429

IvoGoman (approved):

PR #657, opened 2024-10-23
feat(clusterController): check actual token validity

kengou:
## Description

Checking the actual `Bearertoken` validity from `kubeconfig` instead of relying on the cluster status.

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # (issue)
- Closes # (issue)
- Fixes # (issue)

> Remove if not applicable

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist

### DIFF
--- a/pkg/controllers/cluster/util.go
+++ b/pkg/controllers/cluster/util.go
@@ -213,7 +213,6 @@ func reconcileClusterRoleBindingInRemoteCluster(ctx context.Context, k8sClient c
 type tokenHelper struct {
 	client.Client
 	Proxy                              string
-	tokenInfo                          Claims
 	RemoteClusterBearerTokenValidity   time.Duration
 	RenewRemoteClusterBearerTokenAfter time.Duration
 }
@@ -221,6 +220,7 @@ type tokenHelper struct {
 // ReconcileServiceAccountToken reconciles the service account token for the cluster and updates the secret containing the kube config
 func (t *tokenHelper) ReconcileServiceAccountToken(ctx context.Context, restClientGetter *clientutil.RestClientGetter, cluster *greenhousev1alpha1.Cluster) error {
 	var jwtPayload []byte
+	var tokenInfo = new(claims)
 	var actualTokenExpiry metav1.Time
 
 	remoteRestConfig, err := restClientGetter.ToRESTConfig()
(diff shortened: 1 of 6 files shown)


pkg/controllers/cluster/util.go
```diff
@@ -43,6 +62,31 @@ type KubeConfigHelper struct {
 	ClientKeyData  []byte
 }
 
+type Claims struct {
```
IvoGoman, line 65: Only used within this util file.
```suggestion
type claims struct {
```

pkg/controllers/cluster/util.go
```diff
@@ -169,22 +213,47 @@ func reconcileClusterRoleBindingInRemoteCluster(ctx context.Context, k8sClient c
 type tokenHelper struct {
 	client.Client
 	Proxy                              string
+	tokenInfo                          Claims
```
IvoGoman, line 216: This field is filled and accessed only in `ReconcileServiceAccountToken(..)` Could we use a local variable instead if keeping the state is not required?

pkg/controllers/cluster/util.go
```diff
@@ -169,22 +213,47 @@ func reconcileClusterRoleBindingInRemoteCluster(ctx context.Context, k8sClient c
...
+			log.FromContext(ctx).V(5).Info("bearer token is still valid", "cluster", cluster.Name, "expirationTimestamp", cluster.Status.BearerTokenExpirationTimestamp.Time)
+			return nil
+		}
+	}
+
+	actualTokenExpiry = metav1.Unix(t.tokenInfo.Expiry, 0)
+	if actualTokenExpiry.After(time.Now().Add(t.RenewRemoteClusterBearerTokenAfter)) {
+		log.FromContext(ctx).V(5).Info("bearer token is still valid", "cluster", cluster.Name, "expirationTimestamp", cluster.Status.BearerTokenExpirationTimestamp.Time)
+		return nil
+	}
```
IvoGoman, line 255: I think we should skip 251-255 in case the parsing failed and the status does not contain a valid expiration timestamp. In this case the `t.tokenInfo` is initial and we could directly skip to Line 256

IvoGoman (changes requested):

pkg/controllers/cluster/util.go
kengou: sure, will do that

pkg/controllers/cluster/util.go
kengou: fixed in [0ee05f5](https://github.com/cloudoperators/greenhouse/pull/657/commits/0ee05f5db39fba80979132eec273c3e278fdcbaa)

IvoGoman (approved):

PR #658, opened 2024-10-23
feat(greenhousectl) cluster re-onboard updates token

kengou:
## Description

`greenhousectl cluster bootstrap ...` now supports updating tokens for existing clusters. The same command needs to be re-executed then the token will be updated in the `Secret`

Changes in `greenhousectl cluster bootstrap` based on the received feedbacks:
-  `free-access-mode` Flag **retired**
- `bootstrap-kubeconfig` is now `kubeconfig`
- - `--kubeconfig` is expected to point to the cluster which will be onboarded. This parameter does not need to be set.  if `KUBECONFIG` env var is set that will be used
- **NEW FLAG**`--greenhouse-kubeconfig` needs to be set, to point to your greenhouse cluster

`greenhousectl cluster bootstrap` only expects mandatory parameters: `org,  greenhouse-kubeconfig`

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Pre-requisite #657

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code

### DIFF
--- a/pkg/cmd/cluster_bootstrap.go
+++ b/pkg/cmd/cluster_bootstrap.go
@@ -61,16 +61,16 @@ var (
 )
 
 type newClusterBootstrapOptions struct {
-	customerClient     client.Client
-	customerConfig     rest.Config
-	ghClient           client.Client
-	ghConfig           rest.Config
-	kubecontext        string
-	orgName            string
-	clusterName        string
-	customerKubeConfig string
-	onBehafOfUser      string
-	freeAccessMode     bool
+	customerClient       client.Client
+	customerConfig       rest.Config
+	ghClient             client.Client
+	ghConfig             rest.Config
+	kubecontext          string
+	kubeconfig           string
+	orgName              string
+	clusterName          string
+	greenhouseKubeConfig string
+	onBehafOfUser        string
 }
 
 func init() {
@@ -92,6 +92,7 @@ func newClusterBootstrapCmd() *cobra.Command {
 			if err := cmd.ValidateRequiredFlags(); err != nil {
 				return err
 			}
+			o.kubeconfig = cmd.Flag("kubeconfig").Value.String()
 			return o.permissionCheck()
 		},
 	}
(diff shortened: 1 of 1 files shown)


pkg/cmd/cluster_bootstrap.go
```diff
@@ -102,8 +103,7 @@ func newClusterBootstrapCmd() *cobra.Command {
 	bootstrapCmd.Flags().StringVar(&o.kubecontext, "kubecontext", "", "The context to use from the kubeconfig (defaults to current-context)")
```
IvoGoman, line 103: This was used as the `context` from the Greenhouse KubeConfig. Now it is used as the context for the target cluster. The flag description should be explicit about this.

pkg/cmd/cluster_bootstrap.go
```diff
@@ -92,6 +92,7 @@ func newClusterBootstrapCmd() *cobra.Command {
 			if err := cmd.ValidateRequiredFlags(); err != nil {
 				return err
 			}
+			o.kubeconfig = cmd.Flag("kubeconfig").Value.String()
```
IvoGoman, line 95: I am not very familiar with Cobra. This `PersistentPreRunE` is executed before `RunE`, right?
Since the evaluation of the `kubeconfig` flag was moved here, I assume the other flags are evaluated just before run is called? In that do we not need to move the other flags up here as well to ensure the desired values passed as flags are considered?

kengou: Yes It's always executed before the `RunE`, this is why it contains the permission check since the main logic does not even start without the `permissionCheck` to succeed.
`kubeconfig` flag is registered here: https://github.com/kubernetes-sigs/controller-runtime/blob/aaaefb43f7e0e8e1b81371cc1b4705a967dfa0bc/pkg/client/config/config.go#L49-L58 and we are using this function to load the kubeconfig: 
https://github.com/kubernetes-sigs/controller-runtime/blob/aaaefb43f7e0e8e1b81371cc1b4705a967dfa0bc/pkg/client/config/config.go#L116-L136

Since this flag is not registered by Cobra I need to get it somehow to be able to work with that and since `PersistentPreRunE` always runs I load the value of the flag here.

pkg/cmd/cluster_bootstrap.go
kengou: fixed in [b5ada37](https://github.com/cloudoperators/greenhouse/pull/658/commits/b5ada375f35405832d82a8c00aae42bf7e8e30a0)

IvoGoman (approved):

Issue #798, opened 2024-12-11
[FEAT] - Refactor DEX Controller


IvoGoman:
### Priority

(Medium) I'm annoyed but I'll live

### Description

The DexController is responsible for provisioning DEX resources (OAuth2Client, Connector) for each Organization.
Currently these resources are created in the `greenhouse` namespace and are thus not accessible for the Organization.
Furthermore, the OAuth2Clients do not specify a ClientSecret, which is required by the `oauth2-proxy` to be used for securing the services exposed via `service-proxy`.

### Acceptance Criteria

~- [ ] DexController creates the Dex resources in the Organization's namespace~
- [x] Each organization has an additonal OAuth2Client with a generated client secret
- [x] ClientSecret for the OAuth2Client is saved in the organizations namespace secret as `<orgName>-internal`

### Reference Issues

This requires the implementation of #605

abhijith-darshan:
> ### Description
> The DexController is responsible for provisioning DEX resources (OAuth2Client, Connector) for each Organization. Currently these resources are created in the `greenhouse` namespace and are thus not accessible for the Organization. Furthermore, the OAuth2Clients do not specify a ClientSecret, which is required by the `oauth2-proxy` to be used for securing the services exposed via `service-proxy`.
> 
> ### Acceptance Criteria
> * [ ]  DexController creates the Dex resources in the Organization's namespace
> * [ ] OAuth2Clients receive a generated client secret[
> * [ ] ClientSecret for the OAuth2Client is saved in the organizations namespace `<org>-dex-secrets`
> 
> ### Reference Issues
> _No response_

As per the docs dex can only read resources in the namespace it is deployed / running https://dexidp.io/docs/configuration/storage/#configuration

so creating Dex resources in the Organization's namespace would not work

IvoGoman:
I have made changes to the acceptance criteria. This also is now also aligned with the changes of enabling postgres as a backend. When this is enabled there are no kubernetes resources at all

PR #839, opened 2025-01-09
feat(dex): dex backend storage adapter implementation

### LINKED ISSUE #738: [FEAT] - Dex IDP Persistence with DB
### Priority

(Medium) I'm annoyed but I'll live

### User Story

_No response_

### Description

DEX IDP with Kubernetes CRDs as a storage has some instabilities. Look into using a DB as a backend. (rqlite,postgresql) 

rqlite provides helm-chart [repo](https://github.com/rqlite/helm-charts/tree/master/charts/rqlite)

### Acceptance Criteria

- [x] DEX uses a DB instead of Kubernetes CRDs
- [x] Move existing resources to SQL. 
- [ ] DB is deployed from Greenhouse Chart
- [ ] CRDs are cleaned up

### Reference Issues

_No response_

kengou:
## Description

Implements:

- Resolves feature-flags with `dex` storage selection from a `ConfigMap`
- `Dex` backend storage adapter interface to support `kubernetes` or `postgres`
- `idproxy` and `Organization Reconciler` uses the storage adapter interface

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #738 

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📜 README.md

## Checklist


cmd/greenhouse/main.go
```diff
@@ -58,6 +58,9 @@ var (
 	remoteClusterBearerTokenValidity,
 	renewRemoteClusterBearerTokenAfter time.Duration
 	kubeClientOpts clientutil.RuntimeOptions
+	// DB connection parameters
+	pgDB, pgHost, pgUser, pgPasswd string
+	pgPort                         uint16
```
IvoGoman, line 63: WDYT about using the `sql.NetworkDB` struct here. Then we can pass the single struct through to the idproxy/OrgController?

cmd/greenhouse/main.go
```diff
@@ -91,6 +94,12 @@ func main() {
 	flag.StringVar(&common.DNSDomain, "dns-domain", "",
 		"The DNS domain to use for the Greenhouse central cluster")
 
+	flag.StringVar(&pgDB, "database", os.Getenv("DB_NAME"), "Database name")
+	flag.StringVar(&pgHost, "dbHost", os.Getenv("DB_HOST"), "Database host")
+	flag.Uint16Var(&pgPort, "dbPort", 5432, "Database port")
+	flag.StringVar(&pgUser, "dbUser", os.Getenv("DB_USER"), "Database user")
+	flag.StringVar(&pgPasswd, "dbPassword", os.Getenv("DB_PASSWORD"), "Database password")
```
IvoGoman, line 101: The description is quite broad. I think in this case we can make it more specific and prefix with Dex e.g. "Dex Database name". Just to avoid confusion on what the DB is for.

abhijith-darshan: Isn’t it better if these credentials came as secrets instead of args? The deployment can then set it as env variables instead.

abhijith-darshan: Then the connection to DB can be opened during appStart and passed along.

pkg/controllers/organization/dex.go
```diff
@@ -88,6 +95,47 @@ func (r *OrganizationReconciler) reconcileDexConnector(ctx context.Context, org
...
+		Port:     r.PGPort,
+		User:     r.PGUser,
+		Password: r.PGPasswd,
+		Database: r.PGDB,
+	}, logger.With("component", "storage"))
+	if err != nil {
+		return err
+	}
+
+	defer sqlDexStorage.Close()
```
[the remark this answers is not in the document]
kengou: working on it.

cmd/greenhouse/main.go
kengou: done

cmd/greenhouse/main.go
kengou: done, the password would come from a secret, but since the underlying helm chart is not yet uploaded I can not include it here.

pkg/features/features.go
```diff
@@ -0,0 +1,53 @@
...
+	ofinprocess "github.com/open-feature/go-sdk-contrib/providers/go-feature-flag-in-process/pkg"
+	"github.com/open-feature/go-sdk/openfeature"
+	goffclient "github.com/thomaspoignant/go-feature-flag"
+	"github.com/thomaspoignant/go-feature-flag/retriever/k8sretriever"
+	"k8s.io/client-go/rest"
+
+	"github.com/cloudoperators/greenhouse/pkg/clientutil"
+)
+
+func GetOFClient(appName string) *openfeature.Client {
```
IvoGoman, line 20: It would be helpful to have some context to what this client does. Maybe something like this?
```suggestion
// NewClient returns a open-feature client loading feature flags from a ConfigMap
func NewClient(appName string) *openfeature.Client {
```

pkg/features/features.go
```diff
@@ -0,0 +1,53 @@
...
+	k8sInClusterConfig, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatalf("Failed to create in-cluster config: %s", err)
+	}
+
+	goFeatureFlagConfig := &goffclient.Config{
+		PollingInterval: 30 * time.Second,
```
IvoGoman, line 27: Is the polling continuously? In that case we could set the value very high, as this is only evaluated during startup.

cmd/idproxy/main.go
```diff
@@ -59,9 +73,34 @@ func main() {
...
+		if err != nil {
+			log.Fatalf("Failed to initialize postgres storage: %s", err)
+		}
+	case k8sBackend:
+		dexStorage, err = idproxy.NewKubernetesStorage(kubeconfig, kubecontext, kubenamespace, logger.With("component", "storage"))
+		if err != nil {
+			log.Fatalf("Failed to initialize kubernetes storage: %s", err)
+		}
+	default:
+		log.Fatalf("Exactly one of dex-sql-backend or dex-kubernetes-backend must be true")
```
IvoGoman, line 103: We have two flags for the same feature but only one may be toggled. Would it be possible to combine them into one flag with the value for either `k8s` or `postgres`?

pkg/controllers/organization/dex.go
```diff
@@ -88,6 +89,31 @@ func (r *OrganizationReconciler) reconcileDexConnector(ctx context.Context, org
 		log.FromContext(ctx).Info("updated dex connector", "namespace", dexConnector.Namespace, "name", dexConnector.GetName())
 		r.recorder.Eventf(org, corev1.EventTypeNormal, "UpdatedDexConnector", "Updated dex connector %s/%s", dexConnector.Namespace, dexConnector.GetName())
 	}
+
+	// Create the connectors also in SQL storage
+	var oidcConnector storage.Connector
+	if oidcConnector, err = r.dexStorage.GetConnector(org.Name); err != nil {
+		if err = r.dexStorage.CreateConnector(ctx, storage.Connector{
```
IvoGoman, line 96: The error from line 95 must not always be a NotFound error. Directly trying to create a new connector could then lead to AlreadyExists error but obscurring the underlying issue.

pkg/controllers/organization/dex.go
```diff
@@ -118,6 +144,40 @@ func (r *OrganizationReconciler) discoverOIDCRedirectURL(ctx context.Context, or
 }
 
 func (r *OrganizationReconciler) reconcileOAuth2Client(ctx context.Context, org *greenhousesapv1alpha1.Organization) error {
+	var oAuthClient storage.Client
+	var err error
+	if oAuthClient, err = r.dexStorage.GetClient(org.Name); err != nil {
+		if err = r.dexStorage.CreateClient(ctx, storage.Client{
```
IvoGoman, line 150: Same as above w.r.t. obscuring the error condition

cmd/idproxy/main.go
kengou: Not easily with this feature-flag

pkg/features/features.go
kengou: ` 30 * time.Minutes` sounds better ?

pkg/features/features.go
kengou: sure, we are not finished fully with this PR, will do some meaningful commenting into the new functions

pkg/controllers/organization/dex.go
kengou: fixed in https://github.com/cloudoperators/greenhouse/pull/839/commits/74769139e21c0d971151210773dadd82d57324b3

pkg/controllers/organization/dex.go
kengou: fixed in https://github.com/cloudoperators/greenhouse/pull/839/commits/74769139e21c0d971151210773dadd82d57324b3

pkg/features/features.go
kengou: increased in [42393b6](https://github.com/cloudoperators/greenhouse/pull/839/commits/42393b6523a2a77fd4987e959d4c8c0023cea95f)

pkg/features/features.go
kengou: added some comments to the code and renamed the func to better correlate in [4baeee0](https://github.com/cloudoperators/greenhouse/pull/839/commits/4baeee0906ce60627a45e9e085b92cbd28c517fa)

abhijith-darshan:
@IvoGoman @kengou  - Please review from scratch as the implementation has been re-written

Implements:

 - Resolves feature-flags with dex storage selection from a ConfigMap
- Dex backend storage adapter interface to support kubernetes or postgres
- idproxy and Organization Reconciler uses the storage adapter interface
- organization controller will not try to Own dex CRs if the storage is `postgres`, which should allow running both `postgres` and `kubernetes` storage backends in the cluster

Additionally implements #798 -

- If dex storage backend is `kubernetes` then Dex Custom Resources are created in the organization's namespace
- `OAuth2Client` credentials secret is created in the organization's namespace following the pattern `org-name-dex-secrets`
- generates a deterministic client secret using hmac sha256 from clientID and secretKey + truncated to 32 characters (so as to consistently generate the same credential on every reconciliation)

TODO - 

- We need to wrap environment variables in manager deployment and not have a static injection of postgres vars (currently e2e tests and helm lint fail because of that)
- for #798 we need to early exit (requeue?) if the organization `namespace` is not active
- E2E tests for `kubernetes` and `Postgres` should be implemented

pkg/dex/store/k8s_store.go
```diff
@@ -0,0 +1,150 @@
...
+		oAuth2Client.Client.ID = org.Name
+		oAuth2Client.Client.Name = org.Name
+		for _, requiredRedirectURL := range []string{
+			"http://localhost:8085",
+			"http://localhost:8000",
+			"https://dashboard." + common.DNSDomain,
+			fmt.Sprintf("https://%s.dashboard.%s", org.Name, common.DNSDomain),
+		} {
+			oAuth2Client.Client.RedirectURIs = util.AppendStringToSliceIfNotContains(requiredRedirectURL, oAuth2Client.RedirectURIs)
+		}
```
IvoGoman, line 117: I think I remember why we have this in place. This hack was/is used by the UI Team to add additional redirect URIs to the client, to be used when testing the UI against the IDProxy. As noted on the Postgres implementation, this is not easily done there.
With the current implementation it will require manual modification of the entry in the database.

Specifying additional redirectURIs will be required for the `oauth2-proxy` as well, when using it to secure the service proxy. Let's keep the code in and tackle this in an additional issue #893

pkg/features/features.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, err
+	}
+	return &features{
+		raw: featureMap.Data,
+	}, nil
+}
+
+func (f *features) resolveDexFeatures() error {
+	f.m.Lock()
+	defer f.m.Unlock()
```
IvoGoman, line 56: Do we need a mutex here? The map is populated once, during `NewFeatures`. Since the map is a private field on the `features` struct it can't be accessed from anywhere else and from within the package there is just read-only  access from this method.

pkg/features/features.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	f.dex = dex
+	return nil
+}
+
+func (f *features) GetDexStorageType(ctx context.Context) *string {
+	if f.dex != nil {
+		return clientutil.Ptr(f.dex.Storage)
```
IvoGoman, line 77: See above for `k8s.io/utils/ptr`

pkg/features/features.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	if err := f.resolveDexFeatures(); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "failed to resolve dex features")
+		return nil
+	}
+	if f.dex.Storage == "" {
+		return nil
+	}
+	return clientutil.Ptr(f.dex.Storage)
```
IvoGoman, line 86: See above for `k8s.io/utils/ptr`

IvoGoman (changes requested):

pkg/features/features.go
abhijith-darshan: As the feature flags grow there might be a possibility of concurrent access happening from the same controller routine. Hence I thought I add the mutex just in case. If we just look at the single flag that we are currently resolving this is not needed yes. I can add a small comment to be aware of it in the future and remove it.

pkg/dex/web/templates/approval.html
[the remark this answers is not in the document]
abhijith-darshan: Just to keep everything related to dex under one package. Shall I revert this?

pkg/features/features.go
abhijith-darshan: done

pkg/features/features.go
abhijith-darshan: done

pkg/features/features.go
abhijith-darshan: i have removed it as every controller is in it's own routine so mutex here doesn't make sense as it is not a singleton

pkg/dex/store/pg_store.go
```diff
@@ -0,0 +1,150 @@
...
+		authClient.Public = true
+		authClient.ID = org.Name
+		authClient.Name = org.Name
+		for _, requiredRedirectURL := range []string{
+			"http://localhost:8085",
+			"https://dashboard." + common.DNSDomain,
+			fmt.Sprintf("https://%s.dashboard.%s", org.Name, common.DNSDomain),
+		} {
+			authClient.RedirectURIs = util.AppendStringToSliceIfNotContains(requiredRedirectURL, authClient.RedirectURIs)
+		}
```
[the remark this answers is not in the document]
abhijith-darshan: Yes if the DB was updated manually.

However, we also have redirectURI in Org spec as well. This way one org's redirects doesn't need to be in parent greenhouse org redirects.

pkg/features/features.go
```diff
@@ -0,0 +1,87 @@
...
+	dex *dexFeatures `yaml:"dex"`
+}
+
+type dexFeatures struct {
+	Storage string `yaml:"storage"`
+}
+
+type Features interface {
+	GetDexStorageType(ctx context.Context) *string
+}
```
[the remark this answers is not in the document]
abhijith-darshan: we need the interface because 

```go
type features struct {
	raw map[string]string
	dex *dexFeatures `yaml:"dex"`
}

func NewFeatures(ctx context.Context, k8sClient client.Reader, configMapName, namespace string) (Features, error) {...}
```

Now we can add multiple feature structs ex: `cluster *clusterFeatures` and expose only those methods with necessary toggles exposure for ex: `GetDefaultClusterAuthType`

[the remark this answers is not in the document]
abhijith-darshan: done

pkg/clientutil/env.go
```diff
@@ -24,3 +25,10 @@ func GetIntEnvWithDefault(envKey string, def int) int {
 	}
 	return i
 }
+
+func GetEnv(envKey string) (string, error) {
```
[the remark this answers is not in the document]
abhijith-darshan: done

pkg/controllers/plugin/pluginpreset_controller_test.go
```diff
@@ -254,7 +256,7 @@ var _ = Describe("PluginPreset Controller Lifecycle", Ordered, func() {
 			err = test.K8sClient.List(test.Ctx, pluginList, client.InNamespace(cluster.GetNamespace()), client.MatchingLabels{greenhouseapis.LabelKeyPluginPreset: pluginPresetName})
 			g.Expect(err).NotTo(HaveOccurred(), "failed to list Plugins")
 			g.Expect(pluginList.Items).To(HaveLen(1), "there should be only one Plugin")
-		}).Should(Succeed(), "the PluginPreset should have removed the Plugin for the deleted Cluster")
+		}, 2*time.Minute).Should(Succeed(), "the PluginPreset should have removed the Plugin for the deleted Cluster")
```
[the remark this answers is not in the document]
abhijith-darshan: yes seems like it.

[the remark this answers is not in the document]
abhijith-darshan: not here though. It is for test cluster-c `MustDeleteCluster` test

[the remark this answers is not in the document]
abhijith-darshan: The flakiness is being handled by @gciezkowski-acc

pkg/controllers/organization/organization_controller.go
```diff
@@ -111,7 +111,19 @@ func (r *OrganizationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		return ctrl.Result{}, lifecycle.Success, nil
+	}
+
+	if org.Spec.Authentication != nil && org.Spec.Authentication.OIDCConfig != nil {
+		// delete org oauth redirects from default connector
+		if err := r.removeAuthRedirectFromDefaultConnector(ctx, org); err != nil {
+			return ctrl.Result{}, lifecycle.Failed, err
+		}
+	}
+	// if dex storage type is kubernetes then deletion of org connector, org oauth2client will be handled automatically by owner reference
```
IvoGoman, line 126: How are the entries in the database handled?

abhijith-darshan: Hey @IvoGoman on it's way. I had to resolve conflicts before pushing the changes

pkg/controllers/organization/organization_controller.go
```diff
@@ -86,15 +99,39 @@ func (r *OrganizationReconciler) SetupWithManager(name string, mgr ctrl.Manager)
...
+		}
+		// if dex storage type is kubernetes then deletion of org connector, org oauth2client will be handled automatically by owner reference
+		if r.DexStorageType == dexstore.Postgres {
+			if err := r.deleteDexConnector(ctx, org); err != nil {
+				return ctrl.Result{}, lifecycle.Failed, err
+			}
+			if err := r.deleteOAuth2Client(ctx, org); err != nil {
+				return ctrl.Result{}, lifecycle.Failed, err
+			}
+		}
```
IvoGoman, line 133: This delete logic would work for both storage types. What do you think about calling it in both cases?
Then we do not require the manual adding of the owner references on the dex resources in case the kube storage is used (`setDexOwnerReferences(..)`)

abhijith-darshan: Yes that’s cool. I will do the changes.

IvoGoman (dismissed):

IvoGoman (approved):

Issue #838, opened 2025-01-09
🐛 [BUG] - Service Proxy not working for Perses


ibakshay:
### Priority

(Medium) I'm annoyed but I'll live

### service-proxy logs:
```yaml
service-proxy-79677f65cf-4kqx9 2025-01-09T14:40:19Z    INFO    Forwarded request    {"host": "obs-eu-de-1--184931c.ccloud.greenhouse-qa.eu-nl-1.cloud.sap", "url": "/", "method": "GET", "status": 200, "upstream": "https://api.obs-eu-de-1.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/"}
service-proxy-79677f65cf-4kqx9 2025-01-09T14:40:20Z    INFO    Forwarded request    {"host": "obs-eu-de-1--184931c.ccloud.greenhouse-qa.eu-nl-1.cloud.sap", "url": "/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/main.js", "method": "GET", "status": 404, "upstream": "https://api.obs-eu-de-1.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/main.js"}
service-proxy-79677f65cf-4kqx9 2025-01-09T14:40:20Z    INFO    Forwarded request    {"host": "obs-eu-de-1--184931c.ccloud.greenhouse-qa.eu-nl-1.cloud.sap", "url": "/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/favicon.ico", "method": "GET", "status": 404, "upstream": "https://api.obs-eu-de-1.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/api/v1/namespaces/kube-monitoring/services/http:perses:8080/proxy/favicon.ico"}
```

### Perses logs:
```yaml
perses time="2025-01-09T14:16:46Z" level=info func="github.com/perses/common/echo.(*Builder).build.Logger.LoggerWithConfig.func2.1" file="/home/runner/go/pkg/mod/github.com/perses/common@v0.26.0/echo/middleware/logger.go:63" method=GET status=200 uri=/
[...]

PR #868, opened 2025-01-29
fix(service-proxy): avoid Duplicate Proxy Path in URL Rewrite Function

ibakshay:
### **Description**
This PR fixes an issue in the `rewrite` function of the `service-proxy`, where outgoing requests to a upstream  service could contain **duplicate proxy paths**. The issue primarily affected applications like `perses`, where requests for static assets (e.g., JavaScript, CSS, and favicon files) resulted in **404 Not Found** errors due to incorrect URL rewriting.

### **Root Cause**
1. The `rewrite` function previously **unconditionally appended** the Kubernetes API proxy path (`/api/v1/.../proxy/`) to `req.Out.URL.Path`, even when the request already contained the full proxy path.
2. This resulted in URLs like:
   ```
   https://api.blueprints.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/perses:8080/proxy/api/v1/namespaces/kube-monitoring/services/perses:8080/proxy/perses/main.js
   ```
   instead of:
   ```
   https://api.blueprints.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/perses:8080/proxy/perses/main.js
   ```
I have also added a lot more tests to cover more scenarios. 

### **Related Issues**
Closes https://github.com/cloudoperators/greenhouse/issues/838


cmd/service-proxy/proxy_test.go
```diff
@@ -27,63 +27,91 @@ import (
 // If checks if he url is properly rewritten and the request context contains the cluster name
 // and a logger with the correct values.
 func TestRewrite(t *testing.T) {
-	proxyURL, err := url.Parse("https://apiserver/proxy/url")
+	proxyURL, err := url.Parse("https://api.blueprints.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/test-service:8080")
```
abhijith-darshan, line 30: let's avoid real domain names and use mock URLs

cmd/service-proxy/proxy_test.go
```diff
@@ -27,63 +27,91 @@ import (
...
+		contextVal                      any
 	}{
 		{
-			name:        "valid host",
-			url:         "https://cluster--1234567.organisation.basedomain/abcd",
-			expectedURL: "https://apiserver/proxy/url/abcd",
-			contextVal:  "cluster",
+			name:                            "valid host with path",
+			url:                             "https://cluster--1234567.organisation.basedomain/dashboard",
+			expectedupstreamServiceRouteURL: "https://api.blueprints.greenhouse.shoot.canary.k8s-hana.ondemand.com/api/v1/namespaces/kube-monitoring/services/test-service:8080/dashboard",
```
abhijith-darshan, line 44: same here let's avoid real domain names and use mock URLs

cmd/service-proxy/proxy_test.go
```diff
@@ -27,63 +27,91 @@ import (
...
+			name:                            "unknown cluster request",
+			url:                             "https://unknown-cluster.organisation.basedomain/dashboard",
+			expectedupstreamServiceRouteURL: "https://unknown-cluster.organisation.basedomain/dashboard", // No rewrite expected
+			contextVal:                      nil,
+		},
+		{
+			name:                            "invalid host format",
+			url:                             "https://something.organisation.basedomain/abcd",
+			expectedupstreamServiceRouteURL: "https://something.organisation.basedomain/abcd", // No rewrite expected
+			contextVal:                      nil,
```
abhijith-darshan, line 69: let's avoid real domain names and use mock URLs

cmd/service-proxy/proxy.go
```diff
@@ -190,46 +191,69 @@ func (pm *ProxyManager) RoundTrip(req *http.Request) (resp *http.Response, err e
...
-	defer func() {
-		req.Out = req.Out.WithContext(log.IntoContext(req.Out.Context(), l))
-	}()
-
-	// hostname is expected to have the format specified in ./pkg/common/url.go
+	// Extract cluster from the incoming request host
 	cluster, err := common.ExtractCluster(req.In.Host)
 	if err != nil {
+		l.Error(err, "Failed to extract cluster from host", "host", req.In.Host)
+		req.Out = req.Out.WithContext(log.IntoContext(req.Out.Context(), l))
```
abhijith-darshan, line 213: This was handled in defer method to avoid code duplication as you can see it is repeated again in line 221

cmd/service-proxy/proxy.go
```diff
@@ -190,46 +191,69 @@ func (pm *ProxyManager) RoundTrip(req *http.Request) (resp *http.Response, err e
...
 func (pm *ProxyManager) rewrite(req *httputil.ProxyRequest) {
 	req.SetXForwarded()
 
-	l := pm.logger.WithValues("host", req.In.Host, "url", req.In.URL.String(), "method", req.In.Method)
+	// Create a logger with relevant request details
+	l := pm.logger.WithValues(
+		"incomingHost", req.In.Host,
+		"incomingRequestURL", req.In.URL.String(),
+		"incomingMethod", req.In.Method,
+	)
```
abhijith-darshan, line 207: NIT: why don't we use pm.logger for everything instead of using log.FromContext in certain places?

abhijith-darshan (changes requested):
Thank you @ibakshay 🚀 

Some minor feedback

cmd/service-proxy/proxy.go
uwe-mayer: Need to make sure `pm.logger != nil` in all places used. But yes, makes sense!
I think it is initialized in `rewrite` func, which is called first anyhow.

cmd/service-proxy/proxy.go
ibakshay: Thanks a lot. Fixed in 7adddf6a42c1ad616447908c57b07f4bca14601f

cmd/service-proxy/proxy_test.go
ibakshay: sure, good catch. fixed in 7adddf6a42c1ad616447908c57b07f4bca14601f

cmd/service-proxy/proxy.go
ibakshay: Yeah, good catch! That mix of pm.logger and log.FromContext was actually already there before this pull request. I'd say we keep this PR focused on the bug fix, and we can do a quick cleanup/refactoring in a separate PR to make it consistent. Let me know what you think!

cmd/service-proxy/proxy_test.go
ibakshay: sure, good catch. fixed in 7adddf6a42c1ad616447908c57b07f4bca14601f

cmd/service-proxy/proxy_test.go
ibakshay: sure, good catch. fixed in https://github.com/cloudoperators/greenhouse/commit/7adddf6a42c1ad616447908c57b07f4bca14601f

abhijith-darshan (approved):

uwe-mayer (commented):
Thanks @ibakshay for taking this!

Some comments agreeing with @abhijith-darshan  on mock data...

cmd/service-proxy/proxy.go
```diff
@@ -190,46 +191,69 @@ func (pm *ProxyManager) RoundTrip(req *http.Request) (resp *http.Response, err e
...
+		"upstreamServiceRouteURL", req.Out.URL.String(),
+	)
 }
 
 // modifyResponse strips the k8s API server proxy path prepended to the location header during redirects:
 // https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/proxy/transport.go#L113
 func (pm *ProxyManager) modifyResponse(resp *http.Response) error {
+	logger := log.FromContext(resp.Request.Context())
+	logger.Info("Modifying response", "statusCode", resp.StatusCode, "originalLocation", resp.Header.Get("Location"))
+
```
IvoGoman, line 259: I am not very deep into how the whole service proxy works.  The comment on this method indicates that it was originally intended to handle the API server proxy path during redirects. 
cc @uwe-mayer

cmd/service-proxy/proxy.go
```diff
@@ -190,46 +191,69 @@ func (pm *ProxyManager) RoundTrip(req *http.Request) (resp *http.Response, err e
...
-	l.WithValues("cluster", cluster, "namespace", route.namespace, "name", route.serviceName)
+	// Ensure the outgoing request URL is properly updated
+	if !strings.HasPrefix(req.Out.URL.Path, upstreamServiceRouteURL.Path) {
+		// Append the original request path to the upstream service route URL path
+		req.Out.URL.Path = strings.TrimSuffix(upstreamServiceRouteURL.Path, "/") + req.Out.URL.Path
+	}
+
+	// Set the correct upstream service route URL details
+	req.Out.URL.Scheme = upstreamServiceRouteURL.Scheme
+	req.Out.URL.Host = upstreamServiceRouteURL.Host
```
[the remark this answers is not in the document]
ibakshay: the req.SetURL(...) was not handling [this test case (valid host with already prefix path)](https://github.com/cloudoperators/greenhouse/blob/d3d03e766ea2e0ac24236e97a1087f73618c0382/cmd/service-proxy/proxy_test.go#L53-L58) properly. 
Below is the test failure output when we have `req.SetURL(...)`
```bash
expected URL https://api.test-api-server.com/api/v1/namespaces/kube-monitoring/services/test-service:8080/existing-path
got https://api.test-api-server.com/api/v1/namespaces/kube-monitoring/services/test-service:8080/api/v1/namespaces/kube-monitoring/services/test-service:8080/existing-path
```
That is why I had to manually set the `req.out.URL...`.

cmd/service-proxy/proxy.go
ibakshay: >The comment on this method indicates that it was originally intended to handle the API server proxy path during redirects.

Yes, that is correct. I have tested and I can confirm that`modifyResponse` does handle redirects properly without any issues. I have also added tests for this method :)

PR #897, opened 2025-02-11
feat: visualize Greenhouse metrics with Perses

ibakshay:
This PR closes https://github.com/cloudoperators/greenhouse-extensions/issues/618


charts/greenhouse/templates/plugin-perses.yaml
```diff
@@ -0,0 +1,30 @@
...
+          paths:
+            - path: /
+              pathType: Prefix
+    - name: perses.ingress.ingressClassName
+      value: nginx
+    - name: perses.ingress.tls
+      value:
+        - hosts:
+            - {{ index .Values.perses.ingress.tls.hosts 0 }}
+          secretName: {{ printf "%s-%s" "tls" (first .Values.perses.ingress.hosts).host | replace "." "-" }}
```
IvoGoman, line 27: Please have a look at this https://github.com/cloudoperators/greenhouse/blob/6564bc562f1dc56e20d1890c9a3ebf25013ea7dc/charts/idproxy/templates/ingress.yaml#L28-L33
I tried to reduce the amount of values required to be set when deploying Greenhouse itself, and opted to hard-code the subdomains within helm templates. This way it is sufficient to have the one dns domain for Greenhouse and everything else just has to add subdomains.

charts/greenhouse/templates/configmap-perses-dashboards.yaml
```diff
@@ -0,0 +1,17 @@
+{{ if and .Values.plugins.enabled .Values.perses.enabled }}
+{{- range $path, $bytes := .Files.Glob "perses-dashboards/*.json" }}
```
IvoGoman, line 2: This condition should also be around the Plugin, right? I would opt to enable it by default and make it opt-out. Together with the other comment, it would make additional configuration in the internal config repo unnecessary.

ibakshay: Yes, you are right. I have missed it. This is a great feedback.  I have fixed in f3ecefe7cee81372b4b8d25ffd67c9d1591485ea :) I have also closed the PR in the internal repo.

charts/greenhouse/ci/test-values.yaml
```diff
@@ -59,6 +59,15 @@ plutono:
...
       - "foo.bar"
 
+perses:
+  enabled: true
+  ingress:
+    hosts:
+      - host: "foo.bar"
+    tls:
+      hosts:
+        - "foo.bar"
```
IvoGoman, line 69: This is now superfluous with your changes. 
```suggestion
```

charts/greenhouse/templates/plugin-perses.yaml
IvoGoman: Cool. I think for the most part we can get away with some defaults for the ingress hosts. In the end the subdomain can also be made configurable, if the need arises at some point. But I think for now it should be sufficient. Thanks for the changes.

IvoGoman (approved):
Thank you for adding Perses 🚀

PR #923, opened 2025-02-21
feat(greenhouse): test pod logs retrieval

ibakshay:
This PR closes https://github.com/cloudoperators/greenhouse-extensions/issues/589

Signed-off-by: Akshay Iyyadurai Balasundaram <akshay.iyyadurai.balasundaram@sap.com>

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


pkg/controllers/plugin/helm_chart_testing.go
```diff
@@ -93,3 +93,40 @@ func (r *PluginReconciler) reconcileHelmChartTest(ctx context.Context, plugin *g
...
+
+func extractErrorsFromTestPodLogs(testPodLogs string) string {
+	var errors []string
+	var errorBlock strings.Builder
+
+	scanner := bufio.NewScanner(strings.NewReader(testPodLogs))
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		fmt.Println(line)
```
IvoGoman, line 105:
```suggestion
		fmt.Println(line)
```

pkg/controllers/plugin/helm_chart_testing.go
```diff
@@ -93,3 +93,40 @@ func (r *PluginReconciler) reconcileHelmChartTest(ctx context.Context, plugin *g
...
+		case strings.HasPrefix(line, "ok"):
+			// Ignore lines starting with "ok"
+			continue
+		default:
+			if errorBlock.Len() > 0 {
+				// Append additional lines related to the failure
+				errorBlock.WriteString(line + "\n")
+			}
+		}
+	}
```
IvoGoman, line 123: This is specific to the bats framework and how it writes the test logs. While we use it extensively for the Chart Tests of our PluginDefinitions, there is no guarantee every Chart Test will do so.
Chart Tests from embedded upstream charts will also be executed, if it is not possible to disable them via values. 
In case that these test do not use bats, filtering the logs will not work.
To avoid writing the whole pod logs of such a test, I would propose that we only write out the logs in case that these are logs generated by a test using bats.

pkg/controllers/plugin/helm_chart_testing.go
```diff
@@ -93,3 +93,40 @@ func (r *PluginReconciler) reconcileHelmChartTest(ctx context.Context, plugin *g
...
+
+	if errorBlock.Len() > 0 {
+		errors = append(errors, errorBlock.String())
+	}
+
+	// Join the error messages with an empty line separator
+	errorsStr := strings.Join(errors, "\n")
+	return errorsStr
```
IvoGoman, line 131:
```suggestion
	return strings.Join(errors, "\n")
```

pkg/helm/helm.go
```diff
@@ -112,27 +113,58 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
...
-	results, err := rta.Run(plugin.Name)
+
+	testAction := action.NewReleaseTesting(cfg)
+	testAction.Timeout = GetHelmTimeout() // set a timeout for the release testing to avoid waiting forever
+	// Used for fetching logs from test pods
+	testAction.Namespace = plugin.Spec.ReleaseNamespace
+	results, err := testAction.Run(plugin.Name)
 	if err != nil {
-		return hasTestHook, err
+		release, getErr := getLatestRelease(cfg, plugin.Name)
```
IvoGoman, line 129: In this case I think it is fine to just call the action directly, without wrapping it in a function.
```suggestion
		release, getErr := action.NewGet(cfg).Run(releaseName)
```

pkg/helm/helm.go
```diff
@@ -112,27 +113,58 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
...
 	}
 
-	return hasTestHook, nil
+	return hasTestHook, "", nil
+}
+
+func getLatestRelease(cfg *action.Configuration, releaseName string) (*release.Release, error) {
+	getAction := action.NewGet(cfg)
+	return getAction.Run(releaseName)
+}
```
IvoGoman, line 154: See above about directly calling the action
```suggestion
```

pkg/helm/helm.go
```diff
@@ -112,27 +113,58 @@ func InstallOrUpgradeHelmChartFromPlugin(ctx context.Context, local client.Clien
...
+		if getErr != nil {
+			log.FromContext(ctx).Error(getErr, "Failed to get latest release", "plugin", plugin.Name)
+		} else {
+			var err2 error
+			testPodLogs, err2 = printTestPodLogs(ctx, testAction, release)
+			if err2 != nil {
+				log.FromContext(ctx).Error(err2, "Failed to retrieve test pod logs", "plugin", plugin.Name)
+			}
+		}
+		return hasTestHook, testPodLogs, err
```
IvoGoman, line 139: With an early return on error we can remove one level of indentation. The `err` returned from the first function will create a new local variable shadowing the outer scoped error. So we do not overwrite the outer error.

```suggestion
		release, err := getLatestRelease(cfg, plugin.Name)
		if err != nil {
			log.FromContext(ctx).Error(getErr, "Failed to get latest release", "plugin", plugin.Name)
			return hasTestHook, "", err
		}
		testPodLogs, err = printTestPodLogs(ctx, testAction, release)
		if err != nil {
			log.FromContext(ctx).Error(err, "Failed to retrieve test pod logs", "plugin", plugin.Name)
		}
		return hasTestHook, testPodLogs, err
```

IvoGoman (changes requested):

pkg/controllers/plugin/helm_chart_testing.go
ibakshay: Thank you. fixed in 5f3e79a528b9277020d3e55b887bdc4199499bd2

pkg/helm/helm.go
ibakshay: sure. I did in 750e430bf6dcf7ef0628367e1965706330770539 :)

pkg/controllers/plugin/helm_chart_testing.go
ibakshay: Hi @IvoGoman, Thank you for your comment. I made sure that the log filtering is generic and is not specific to bats framework. For example, we are not using bats framework for Plutono plugin and you can see how the failed plutono test pod is stored in plutono plugin status field. 

```yaml
apiVersion: greenhouse.sap/v1alpha1
kind: Plugin
metadata:
  name: plutono
  namespace: demo
spec:
  ...
  ...
status:
  statusConditions:
    conditions:
    - lastTransitionTime: "2025-03-04T10:40:26Z"
      message: |+
        not ok 1 Test Health
        # (in test file /tests/run.sh, line 5)
        #   `[ "$code" == "200" ]' failed

      status: "False"
      type: HelmChartTestSucceeded
    - lastTransitionTime: "2025-02-28T14:28:44Z"
      message: Workload is running
      status: "True"
      type: WorkloadReady
  version: 0.1.12

pkg/helm/helm.go
ibakshay: Thank you very much for the suggestion. Fixed in bffcd5e76e2a5da8004984a7bebd767c68546091

pkg/controllers/plugin/helm_chart_testing.go
IvoGoman: The Plutono Plugin does use bats: https://github.com/cloudoperators/greenhouse-extensions/blob/main/plutono/charts/templates/tests/test-configmap.yaml#L14-L18

The point is merely that this filtering is specific to having `not ok` or `ok` in the container logs. If this is not the case the whole output of the test container will be written to the status. I am not sure that is what we want.

ibakshay: Ahhh now I get it. I have fixed in this commit 2f8767fe68cfde5379cb7dd64dccc3110800e608 and I have also added additonal test cases :) Thank you for the suggestion!

IvoGoman (approved):
Thanks 🚀

richardtief:
🚀 :tes

PR #927, opened 2025-02-21
feat(teamrolebindings): teamrbac supports single users

### LINKED ISSUE #920: [FEAT] - TeamRBAC to support single users
### Priority

None

### User Story

> As an org admin I would like to give permissions to single users, so that I can give permissions to technical users which are not part of any groups.

### Description

Managing RBAC with TeamRoleBindings works for users that belong to a team. Sometimes this is not possible for example technical users may not be part of any teams in Greenhouse.

This issue should address this by allowing to specify a list of usernames on a TeamRoleBinding. The controller should then create ClusterRoleBindings and RoleBindings with additional subjects of type:

```yaml
subjects:
- kind: User
  name: foobar # Name is case sensitive
  apiGroup: rbac.authorization.k8s.io
```

Additionally, there should be an optional field `createNamespaces` on the TeamRoleBindings to create namespaces if they did not previously exists. This is to prevent AuthZ issues if a technical users requires access to create resources in a namespace, but the namespace does not exist. Also this will ensure RBAC is rolled out as desired, even if the Plugins that create the namespace will be rolled out at a later stage.
The default should however be `createNamespaces = False`

### Acceptance Criteria

- [ ] Users can be specified on the TeamRoleBinding
- [ ] `createNamespaces` can be specified on a TeamRoleBinding to ensure namespaces are created 

### Reference Issues

_No response_

gciezkowski-acc:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # (issue)
- Closes #920 
- Fixes # (issue)

## Added tests?

- [X] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


pkg/controllers/teamrbac/teamrolebinding_controller_test.go
```diff
@@ -386,6 +390,26 @@ var _ = Describe("Validate ClusterRole & RoleBinding on Remote Cluster", Ordered
...
+				test.WithTeamRoleRef(teamRoleUT.Name),
+				test.WithTeamRef(teamUT.Name),
+				test.WithClusterName(clusterA.Name),
+				test.WithNamespaces("non-existing-namespace"),
+				test.WithCreateNamespace(true))
+
+			By("checking that the Namespace is created")
+			namespace := &corev1.Namespace{}
+			Eventually(func(g Gomega) {
+				err := test.K8sClient.Get(test.Ctx, types.NamespacedName{Name: "non-existing-namespace"}, namespace)
```
IvoGoman, line 406: The used client is for the greenhouse cluster, this test should verify that the namespace was created on the remote cluster.

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -463,16 +467,27 @@ func rbacClusterRoleBinding(trb *greenhousev1alpha1.TeamRoleBinding, clusterRole
...
-				APIGroup: rbacv1.GroupName,
-				Kind:     rbacv1.GroupKind,
-				Name:     team.Spec.MappedIDPGroup,
-			},
-		},
+		Subjects: joinUsernamesWithDefault(trb.Spec.Usernames, team.Spec.MappedIDPGroup),
 	}
 }
 
+func joinUsernamesWithDefault(usernames []string, mappedIDPGroup string) []rbacv1.Subject {
```
IvoGoman, line 474:
```suggestion
// generateSubjects returns a list of subjects with mappedIDPGroup as a rbacv1.GroupKind, and any usernames as rbacv1.UserKind
func generateSubjects(usernames []string, mappedIDPGroup string) []rbacv1.Subject {
```

IvoGoman (changes requested):

pkg/apis/greenhouse/v1alpha1/teamrolebinding_types.go
```diff
@@ -17,13 +17,18 @@ type TeamRoleBindingSpec struct {
 	TeamRoleRef string `json:"teamRoleRef,omitempty"`
 	// TeamRef references a Greenhouse Team by name
 	TeamRef string `json:"teamRef,omitempty"`
+	// Usernames define list of users with team role
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/apis/greenhouse/v1alpha1/teamrolebinding_types.go
```diff
@@ -17,13 +17,18 @@ type TeamRoleBindingSpec struct {
...
+	Usernames []string `json:"usernames,omitempty"`
 	// ClusterName is the name of the cluster the rbacv1 resources are created on.
 	ClusterName string `json:"clusterName,omitempty"`
 	// ClusterSelector is a label selector to select the Clusters the TeamRoleBinding should be deployed to.
 	ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty"`
 	// Namespaces is a list of namespaces in the Greenhouse Clusters to apply the RoleBinding to.
 	// If empty, a ClusterRoleBinding will be created on the remote cluster, otherwise a RoleBinding per namespace.
 	Namespaces []string `json:"namespaces,omitempty"`
+	// CreateNamespaces when the flag is set the controller will create namespaces defined in the Namespaces field.
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -48,6 +48,7 @@ type TeamRoleBindingReconciler struct {
 //+kubebuilder:rbac:groups=greenhouse.sap,resources=teamrolebindings/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=greenhouse.sap,resources=teamrolebindings/finalizers,verbs=update
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=create;get
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -785,3 +800,23 @@ func computeReadyCondition(status greenhousev1alpha1.TeamRoleBindingStatus) gree
...
+			err := r.Client.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: nampespaceName}})
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -549,10 +542,15 @@ func reconcileClusterRoleBinding(ctx context.Context, cl client.Client, c *green
 		remoteCRB.Subjects = crb.Subjects
 		return nil
 	})
+	if err != nil {
+		return err
+	}
 
+	err = createNamespaces(ctx, trb, cl)
```
IvoGoman, line 548: The `reconcileClusterRoleBinding` is called *only* when there are no namespaces set on the ClusterRoleBindings. The ClusterRoleBindings that are created by this method are cluster scoped resources. It will never be the case that the namespace of the TeamRoleBinding will be filled.

IvoGoman: This logic will also not work, as the namespace needs to be created before the rolebinding can be created.
It would require something like 

```golang
if err != nil{
 if apierrors.IsNotFound(err){
  // create the namespace for the RoleBinding that is currently reconciled
}
return err
}

```

IvoGoman (changes requested):

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -233,6 +233,12 @@ func (r *TeamRoleBindingReconciler) doReconcile(ctx context.Context, teamRole *g
 			trb.SetPropagationStatus(cluster.GetName(), metav1.ConditionTrue, greenhousev1alpha1.RBACReconciled, "")
 		default:
 			errorMesages := []string{}
+			err = createNamespaces(ctx, trb, remoteRestClient)
+			if err != nil {
+				trb.SetPropagationStatus(cluster.GetName(), metav1.ConditionFalse, greenhousev1alpha1.CreateNamespacesFailed, "Failed to create namespaces: "+err.Error())
+				continue
+			}
```
IvoGoman, line 240: This will try to create the namespaces on every reconciliation of the TeamRoleBinding. This will result in at least one get request per namespace.
Please move the creation of the individual namespaces into this part of the creation of the rolebinding on the remote cluster https://github.com/cloudoperators/greenhouse/blob/2827d24fed1a402a9686c90be446bebaa952ba8b/pkg/controllers/teamrbac/teamrolebinding_controller.go#L568-L570

The create of a RoleBinding will fail with an `NotFound` error if the namespace does not exists.
That error can be caught as previously suggested with `apierrors.IsNotFound(..)`. 

In case it is such an error the namespace can be created. It should still return an error so that the TeamRoleBinding is reconciled once more. In the second reconciliation the create should succeed since the namespace was created.

something like
```golang
if err != nil{
 if apierrors.IsNotFound(err){
  // create the namespace for the RoleBinding that is currently reconciled
}
return err
}
```

IvoGoman (changes requested):

pkg/controllers/teamrbac/teamrolebinding_controller.go
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
gciezkowski-acc: Done

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -550,24 +558,39 @@ func reconcileClusterRoleBinding(ctx context.Context, cl client.Client, c *green
...
 	result, err := clientutil.CreateOrPatch(ctx, cl, remoteRB, func() error {
 		remoteRB.Labels = rb.Labels
 		remoteRB.RoleRef = rb.RoleRef
 		remoteRB.Subjects = rb.Subjects
 		return nil
 	})
-
 	if err != nil {
 		return err
 	}
```
IvoGoman, line 592: This is what I meant before. If we move the check for the `apierrors.IsNotFound` into the error handling for the CreateOrPatch, then it will fail the first reconciliation if the namespace does not exist. But since if the createNamespace is true, then it will create the namespace during error handling and succeed on the next reconciliation.
Doing it this way there is no need to do the `Get(..)` on the namespace on every iteration.  
```suggestion
	if err != nil {
		if apierrors.IsNotFound(err) && createNamespaces {
			err := cl.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: rb.Namespace}})
			if err != nil {
				return err
			}
			return errors.New("failed to create RoleBinding, namespace did not exist")
		}
	return err
	}
```

IvoGoman (changes requested):

pkg/controllers/teamrbac/teamrolebinding_controller.go
```diff
@@ -564,10 +572,17 @@ func reconcileRoleBinding(ctx context.Context, cl client.Client, c *greenhousev1
...
 		return nil
 	})
-
 	if err != nil {
-		return err
+		if createNamespaces && apierrors.IsNotFound(err) {
+			err := cl.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: rb.Namespace}})
+			if err != nil {
+				return err
+			}
```
IvoGoman, line 580: An error has to be returned in every case to force the reconciliation to be triggered again.
```suggestion
			err := cl.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: rb.Namespace}})
			if err != nil {
				return err
			}
			return fmt.Errorf("failed to create RoleBinding, created missing namespace")
```

IvoGoman (approved):

PR #1104, opened 2025-04-29
chore(greenhouse): use generateName for one time jobs

abhijith-darshan:
## Description

- use generate name for one time jobs.

One time jobs self-deletes after successful completion., but GitOps tools like Argo tend to re-apply the resource infinitely as it sees the resource to be maintained declaratively in git or helm chart.

To prevent this from infinite replacement, the job can be annotated with -

```
  annotations:
    argocd.argoproj.io/hook: PostSync
    argocd.argoproj.io/hook-delete-policy: HookSucceeded
```

This will prevent Argo from re-applying after successful completion.

However if we would want the job to run on every argocd sync we should use `generateName` instead of fixed names.

see [argocd resource hooks](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/#generate-name)

for greenhouse it will be better to run the `kube-webhook-cert-gen` job on every deployment....

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


abhijith-darshan:
closing this as it is not needed for now...

Issue #892, opened 2025-02-10
[EPIC] - IDProxy stability


IvoGoman:
### Description

The `idproxy` uses dexidp under the hood to allow JTWIssuer multiplexing in the Greenhouse APIServer. 
There are currently some issues with stability when using etcd as the storage backend for DEX.

### Objectives

- Replace the Storage Backend used by Dex with Postgres
- Make the OAuth2Client credentials available in the organisation namespace
- Allow to specify additional redirectURL allowed by the IDProxy

### Acceptance Criteria

- [x] Dex is using postgres as the backend
- [x] oauth2-proxy can consume the OAuth2Client credentials from the org namepace
- [x] additional redirect URL for oauth2-proxy can be defined 

### Dependencies

- Dependency 1
- Dependency 2
- Dependency 3

PR #1105, opened 2025-04-29
feat(service-proxy): authenticated service-proxy endpoints via dex

### LINKED ISSUE #701: [FEAT] - Secure resources exposed urls with the `service-proxy`
### User Story

> As a Greenhouse Admin I want to restrict access to URLs exposed with the `service-proxy`, so that the audience of a service can be controlled.

A second use case is to allow access to exposed services for technical users (e.g. with api key or tls certificate) 

### Description

As is everyone who has access to a URL exposed via the `service-proxy` can access it. In order to avoid misuse of this feature and the exposed service the access should be restricted. This is a PCI compliance issue as access to all administrative UIs (even read-only UIs like prometheus/plutono) has to be protected.

### Solution Proposal
For authentication we will reuse the `idproxy` which is already used to protect the greenhouse UI.
This will make sure that only users with access the the organization can assess the exposed services.
The service proxy is currently exposed via the nginx-ingress and a [single org-wide ingress resource](https://github.com/sapcc/helm-charts/blob/master/global/oauth-proxy/templates/ingress.yaml) with a wildcard certificate.
This means we can use a similar approach as we do for elektra: We deploy an oauth2-proxy next the service proxy in each organization and configure it to use the `idproxy` as an upstream oidc authentication provider.
By adding [an annotation](https://kubernetes.github.io/ingress-nginx/examples/auth/oauth-external-auth/) the the service-proxy ingress definition we can make sure the ingress will only forward authenticated requests to the service-proxy instances.

Pros:
* Re-use of proven implementation (elektra), [oauth-proxy chart](https://github.com/sapcc/helm-charts/tree/master/global/oauth-proxy)
* Low implementation effort, mostly adding resources to the [service-proxy chart](https://github.com/cloudoperators/greenhouse-extensions/tree/main/service-proxy)

Cons:
* Only authentication, no fine grained authorization possible per exposed-service

For the second use case we could a second ingress resource that requires client tls credentials instead of oidc authentication. The client certificate should be manageble by org admins (e.g. in a secret)

### Acceptance Criteria

- [x] service-proxy URLs should be protected automatically
- [x] by default all Organization members should have access to the `service-proxy` URLs
- [ ] Support for certificates for accessing exposed endpoints interactively 

### Reference Issues

_No response_

### LINKED ISSUE #798: [FEAT] - Refactor DEX Controller

abhijith-darshan:
## Description

This PR implements -

- `service-proxy` endpoints to be authenticated using `oauth2-proxy` via `Dex`

This provides convenience by configuring a one-time dex authentication callback URL in upstream IDP

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue #892 
- Closes #798 
- Closes #701 

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


internal/controller/organization/service_proxy.go
```diff
@@ -28,12 +31,101 @@ import (
...
+		}
+		intSecret.Data[cookieSecretKey] = []byte(cookieData)
+	}
+	oAuthProxyClientName := fmt.Sprintf("%s-%s", dexOAuth2ProxyClientPrefix, org.Name)
+	intSecret.Data[dexOAuth2ProxyClientIDKey] = []byte(oAuthProxyClientName)
+	oauthProxyClientSecret, err := generateOauth2ProxySecret()
+	if err != nil {
+		log.FromContext(ctx).Error(err, "failed to create oauth2 proxy client secret for org", "name", org.Name)
+		return err
+	}
```
IvoGoman, line 81: Should this move into the if condition below, so we only generate a new one if it does not yet exists?

internal/controller/organization/service_proxy.go
```diff
@@ -112,31 +177,51 @@ func (r *OrganizationReconciler) reconcileServiceProxy(ctx context.Context, org
...
+					}
+					gardenAnnBytes, err := json.Marshal(gardenAnnotations)
+					if err != nil {
+						return fmt.Errorf("failed to marshal garden annotations: %w", err)
+					}
+					ingressOptionValue := greenhousev1alpha1.PluginOptionValue{
+						Name:  "ingress.annotations",
+						Value: &apiextensionsv1.JSON{Raw: gardenAnnBytes},
+					}
+		*/
```
IvoGoman, line 197: Good point. We should keep in mind that we likely need to provide a mechanism to overwrite values in these auto-provisioned plugins in the future.

Let's track this https://github.com/cloudoperators/greenhouse/issues/1143, but remove the commented code here.

internal/controller/organization/service_proxy_test.go
```diff
@@ -99,7 +99,27 @@ var _ = Describe("Organization ServiceProxyReconciler", Ordered, func() {
...
+			By("creating greenhouse organization with OIDC config")
+			defaultOrg := setup.CreateOrganization(test.Ctx, "greenhouse", func(org *greenhousev1alpha1.Organization) {
+				org.Spec.MappedOrgAdminIDPGroup = "SOME_IDP_GROUP"
+			})
+			test.EventuallyCreated(test.Ctx, test.K8sClient, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: defaultOrg.Name}})
+			By("creating a secret with dummy oidc config for default organization")
+			secret := setup.CreateSecret(test.Ctx, "greenhouse-oidc-config", test.WithSecretNamespace(defaultOrg.Name), test.WithSecretData(map[string][]byte{"clientID": []byte("dummy"), "clientSecret": []byte("top-secret")}))
+
+			By("Updating the default organization with oidc config")
+			defaultOrg = setup.UpdateOrganization(test.Ctx, defaultOrg.Name, test.WithOIDCConfig("some-issuer.tld", secret.Name, "clientID", "clientSecret"))
```
IvoGoman, line 113: Good one, this would be another one to gradually reduce the repetitive code for updating the Greenhouse resources across our tests 👍

internal/controller/organization/service_proxy.go
```diff
@@ -28,12 +31,101 @@ import (
...
+	oAuthProxyClient, err := r.dex.GetClient(oAuthProxyClientName)
+	oAuthProxyCallbackURL := fmt.Sprintf("https://auth-proxy.%s/oauth2/callback", getOauthProxyURL(org.Name))
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			if err = r.dex.CreateClient(ctx, storage.Client{
+				Public:       false,
+				ID:           oAuthProxyClientName,
+				Name:         org.Name + " Service Proxy",
+				RedirectURIs: []string{oAuthProxyCallbackURL}, // add service proxy redirect URI
+				Secret:       oauthProxyClientSecret,
```
IvoGoman, line 96: I think we should use `intSecret.Data[dexOAuth2ProxyClientSecretKey]` here to avoid any skew between the client-secret being used by the dex.oauth2client and the secret. If the initial create of the dex.oauth2client fails it will use the newly generated one instead of the one stored in the secret.

abhijith-darshan: done... I seem to have missed this in the `CreateClient` as I had kept what you suggested in `UpdateClient`

internal/controller/organization/service_proxy.go
abhijith-darshan: done

internal/controller/organization/service_proxy.go
abhijith-darshan: done

IvoGoman (approved):

PR #1140, opened 2025-05-14
feat(clusters): propgate lables from cluster secret to cluster

### LINKED ISSUE #1070: [FEAT] - Transport labels from cluster secret to cluster resource
### Priority

None

### User Story

> As a cluster-admin I will most likely maintain the secret resources representing the clusters. If I label a secret I want to have the same labels on the resulting cluster.
> This also enables ownership of clusters via labeling on the secret.

### Description

The [bootstrap controller](https://github.com/cloudoperators/greenhouse/blob/main/internal/controller/cluster/bootstrap_controller.go) should transport labels on the secret to the cluster resource.

### Acceptance Criteria

- [ ] When a secret for bootstraping a cluster via kubeconfig or via OIDC is created, the labels on the secret are present on the bootstrapped clusert
- [ ] Code is validated via tests

### Reference Issues

_No response_

abhijith-darshan:
## Description

- propagates labels from a defined set of label keys in source resource annotation to the destination resource
- maintains propagated labels from src to dest as `last-applied-propagator` annotation in destination resource for tracking and comparison

example:

**cluster secret**:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: kind-greenhouse-remote
  namespace: demo
  labels:
    test-label: test-val
    region: bar
    support_group: foo
  annotations:
    greenhouse.sap/propagate-labels: "support_group, region"
...
```

After label propagation 

**cluster CR**:

```yaml
apiVersion: greenhouse.sap/v1alpha1
kind: Cluster
metadata:
  annotations:
    greenhouse.sap/cluster-connectivity: kubeconfig
    greenhouse.sap/last-applied-propagator: '{"labelKeys":["support_group","region"]}'
  labels:
    region: bar
    support_group: foo
...
```

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #1070 

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


internal/lifecycle/propagation.go
```diff
@@ -0,0 +1,186 @@
...
+	annotationVal := strings.TrimSpace(p.src.GetAnnotations()[PropagateLabelsAnnotation])
+	if annotationVal == "" {
+		return nil
+	}
+	var declared struct {
+		Keys []string `json:"keys"`
+	}
+	if err := json.Unmarshal([]byte(annotationVal), &declared); err != nil || len(declared.Keys) == 0 {
+		return nil
+	}
```
IvoGoman, line 90: What is the reasoning of this nested structure in favour of a shallow list of keys?

internal/lifecycle/propagation.go
```diff
@@ -0,0 +1,186 @@
...
+	return declared.Keys
+}
+
+// hasAtLeastOneValidSourceLabel - returns true if the source object contains
+// at least one of the label keys declared for propagation.
+func (p *Propagator) hasAtLeastOneValidSourceLabel(keys []string, srcLabels map[string]string) bool {
```
IvoGoman, line 96:
```suggestion
func (p *Propagator) containsLabelToPropagate(keys []string, srcLabels map[string]string) bool {
```

IvoGoman (changes requested):
When reading through the code I was not immediately sure from the method names if the operation was on source or target object. Tried to leave some suggestions, not sure if they are ideal either.

w.r.t. to the structure, since this is explicit to annotations I think a flat list should be sufficient to keep the names of the labels to propagate.

internal/lifecycle/propagation.go
abhijith-darshan: @IvoGoman - I wanted to give the option of propagating annotations in the future as well if needed. And hence wanted a JSON structure so we can call it as labelKeys & annotationKeys or something similar or even enhance it with some other conditional properties.

IvoGoman: I think there_ is no current requirement to have this, I would keep it simple and we can always enhance this once we need it. The annotation is currently called `greenhouse.sap/propagate-labels`, so we could then for a more generic version call it `greenhouse.sap/propagate-metadata` or something else.

internal/lifecycle/propagation.go
```diff
@@ -75,25 +74,27 @@ func (p *Propagator) ApplyLabels() client.Object {
...
-// extractDeclaredLabelKeys - retrieves the list of label keys from the propagate-labels annotation
+// labelKeysToPropagate - retrieves the list of label keys from the propagate-labels annotation
 // in the source object. Returns nil if missing, invalid, or empty.
-func (p *Propagator) extractDeclaredLabelKeys() []string {
-	annotationVal := strings.TrimSpace(p.src.GetAnnotations()[PropagateLabelsAnnotation])
-	if annotationVal == "" {
+func (p *Propagator) labelKeysToPropagate() []string {
+	var keys []string
+	annotation := strings.TrimSpace(p.src.GetAnnotations()[PropagateLabelsAnnotation])
+	if strings.TrimSpace(annotation) == "" {
```
IvoGoman, line 82: If the annotation is an empty string the previous call to strings.TrimSpace should have cleared any whitespaces. The second call should have no effect, I think
```suggestion
	annotation := strings.TrimSpace(p.src.GetAnnotations()[PropagateLabelsAnnotation])
	if annotation == "" {
```

internal/controller/organization/organization_controller_test.go
```diff
@@ -341,38 +344,38 @@ var _ = Describe("Test Organization reconciliation", Ordered, func() {
...
 				By("checking if the dex resources are deleted")
-				err = setup.List(test.Ctx, connectors)
-				Expect(err).ToNot(HaveOccurred(), "there should be no error listing dex connectors")
-				Expect(connectors.Items).To(BeEmpty(), "there should be no dex connector resources")
 				err = setup.List(test.Ctx, oAuthClients)
 				Expect(err).ToNot(HaveOccurred(), "there should be no error listing dex oauth clients")
-				Expect(oAuthClients.Items).To(BeEmpty(), "there should be no dex oauth clients resources")
+				for _, orgClient := range oAuthClients.Items {
+					Expect(orgClient.ID).ToNot(Equal(oidcOrg.Name), "the oauth client ID should be equal to organization name")
+				}
```
IvoGoman, line 378: NIT: Could do this with gomega. The expectation message seems off. I think it should be sth. `the oauth client ID of the deleted org should not exist`?
```suggestion
		                Expect(oAuthClients.Items).ToNot(ContainElement(
					MatchFields(IgnoreExtras, Fields{"ID": Equal(oidcOrg.Name)}),
				), "the oauth client list should not contain the deleted organization")
```

IvoGoman (approved):

PR #1227, opened 2025-06-23
feat(ISSUE-978): support group validating webhook

### LINKED ISSUE #978: [FEAT] - support_group validating webhook
### Priority

None

### User Story

> When creating a resource with the `greenhouse.sap/support-group` label, we want to make sure the provided value in the label is an actual support group.

### Description

Within an [validating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) we want to make sure that:
1. the `greenhouse.sap/support-group` label is present on all `PluginPresets`, `Plugins`, `Clusters` and `Secrets`
2. the value of `greenhouse.sap/support-group` is the name of an existing `Team` with the label `greenhouse.sap/support-group: true`

### Acceptance Criteria

- [ ] Creation / Update of `PluginPresets`, `Plugins`, `Clusters` and `Secrets` without `greenhouse.sap/support-group` label is denied
- [ ] Label value of `greenhouse.sap/support-group` is the name of a `Team` labeled with `greenhouse.sap/support-group:true`
- [ ] Behavior is validated through tests

### Reference Issues

_No response_

Zaggy21:
## Description

This PR adds a check for greenhouse.sap/owned-by label to the webhooks of PluginPresets, Plugins, Clusters, TeamRoleBindings, Secrets (only custom Greenhouse ones). If the label is missing or is not set to an existing support-group Team, then it returns an admission warning (but does not deny the creation/update).
It adds an OwnerLabelSet Status Condition set in controllers for PluginPresets, Plugins, Clusters, TeamRoleBindings.
All the tests have been adapted to not throw the warning.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #978 

## Added tests?

- [x] 👍 yes

All the tests have been adapted to not throw the warning introduced by this PR.

## Added to documentation?

- [ ] 📜 README.md

## Checklist


internal/test/resources.go
```diff
@@ -294,6 +314,49 @@ func NewPlugin(ctx context.Context, name, namespace string, opts ...func(*greenh
...
+
+// WithPluginPresetPluginSpec sets the PluginSpec on a PluginPreset.
+func WithPluginPresetPluginSpec(pluginSpec greenhousev1alpha1.PluginSpec) func(*greenhousev1alpha1.PluginPreset) {
+	return func(pp *greenhousev1alpha1.PluginPreset) {
+		pp.Spec.Plugin = pluginSpec
+	}
+}
+
+// WithPluginPresetOwnedByLabelValue sets the value of the greenhouseapis.LabelKeyOwnedBy label on a PluginPreset
+func WithPluginPresetOwnedByLabelValue(value string) func(*greenhousev1alpha1.PluginPreset) {
```
mikolajkucinski, line 332: nitpick: instead of writing separate helper functions for each label, could we generalize this to accept a `map[string]string` of labels?
That way, adding new labels in the future doesn’t require new functions—just pass the extra key/values into this one.

internal/util/conditions.go
```diff
@@ -0,0 +1,69 @@
...
+	if namespace == "" {
+		ownerLabelSetCondition.Message = "Resource namespace is required to validate the owner"
+		ownerLabelSetCondition.Status = metav1.ConditionFalse
+		return ownerLabelSetCondition
+	}
+
+	resourceLabels := resourceObj.GetLabels()
+
+	ownerName, ok := resourceLabels[greenhouseapis.LabelKeyOwnedBy]
+	if !ok {
```
mikolajkucinski, line 32: Since we are setting the same `Reason` and only the `Message` differs a little bit then maybe we could simply put

```go
if !ok || ownerName == "" {
		ownerLabelSetCondition.Reason = greenhousemetav1alpha1.OwnerLabelMissingReason
		ownerLabelSetCondition.Message = fmt.Sprintf("Label %s is missing", greenhouseapis.LabelKeyOwnedBy)
		ownerLabelSetCondition.Status = metav1.ConditionFalse
}

```

internal/webhook/v1alpha1/secret_webhook.go
```diff
@@ -55,10 +63,19 @@ func ValidateCreateSecret(ctx context.Context, _ client.Client, o runtime.Object
 	if err := validateSecretGreenHouseType(ctx, secret); err != nil {
 		return nil, err
 	}
-	return nil, validateKubeconfigInSecret(secret)
+	if err := validateKubeconfigInSecret(secret); err != nil {
+		return nil, err
+	}
+	if slices.Contains(greenhouseSecretTypes, secret.Type) {
```
IvoGoman, line 69: Since we only care about secrets with the special types in this webhook, we can do an early-return in the beginning if the condition is not fulfilled. `!slices.Contains(greenhouseSecretTypes, secret.Type)`
This way the other checks are not called and the label validation can be done for all secrets. 

Same for the ValidateUpdateSecret.

internal/test/resources.go
Zaggy21: That's true, thanks. I added it to simplify refactoring all the tests, but now I can easily change it.

internal/webhook/v1alpha1/secret_webhook.go
Zaggy21: Done.

internal/util/conditions.go
```diff
@@ -16,36 +16,26 @@ import (
...
-	resourceLabels := resourceObj.GetLabels()
-
-	ownerName, ok := resourceLabels[greenhouseapis.LabelKeyOwnedBy]
-	if !ok {
+	ownerName, ok := resourceObj.GetLabels()[greenhouseapis.LabelKeyOwnedBy]
+	if !ok || ownerName == "" {
 		ownerLabelSetCondition.Reason = greenhousemetav1alpha1.OwnerLabelMissingReason
 		ownerLabelSetCondition.Message = fmt.Sprintf("Label %s is missing", greenhouseapis.LabelKeyOwnedBy)
 		ownerLabelSetCondition.Status = metav1.ConditionFalse
 		return ownerLabelSetCondition
```
IvoGoman, line 32: Instead of passing the currently set Condition, we can return the calculated one as such:
```suggestion
		return greenhousemetav1alpha1.FalseCondition(greenhousemetav1alpha1.OwnerLabelSetCondition, greenhousemetav1alpha1.OwnerLabelMissingReason, fmt.Sprintf("Label %s is missing", greenhouseapis.LabelKeyOwnedBy))
```

This also ensures no Reason/Message is carried over from another Condition. e.g. [no reason set](https://github.com/cloudoperators/greenhouse/pull/1227/files/5e2164c3b714404efc008c472d817808cf4ed97d..f6f17f5de9fa9a1bfef49dbe14a6ac374ec1c8e1#diff-08e24ac4618e1fa125198c83c51d8eb5f4df8d921c5d4c20df9f7c3569d386cbR22-R24), [no reason/message](https://github.com/cloudoperators/greenhouse/pull/1227/files/5e2164c3b714404efc008c472d817808cf4ed97d..f6f17f5de9fa9a1bfef49dbe14a6ac374ec1c8e1#diff-08e24ac4618e1fa125198c83c51d8eb5f4df8d921c5d4c20df9f7c3569d386cbR57) set

The condition should be set to unknown by the controller during the initial reconciliation. Every path within `ComputeOwnerLabelCondition` leads to returning the condition set to either True or False.
The `SetConditions` will make sure that the timestamp stays the same, as long as the condition returned from here is idempotent. 

This way we can also remove the `ownerLabelSetCondition` and do not need to retrieve the `greenhousemetav1alpha1.OwnerLabelSetCondition` e.g. [here](https://github.com/cloudoperators/greenhouse/pull/1227/files/5e2164c3b714404efc008c472d817808cf4ed97d..f6f17f5de9fa9a1bfef49dbe14a6ac374ec1c8e1#diff-a2320023a83daf79beae4ecdea5b77e328c282f80bd8f0e99ab87b76f04e203eR98-R99)

IvoGoman (changes requested):

internal/util/conditions.go
Zaggy21: Thanks for clarifying that up. I must have got confused with how the StatusConditions work.

Issue #1000, opened 2025-03-17
[FEAT] - Controller for Catalogs implemented


IvoGoman:
### Priority

(Medium) I'm annoyed but I'll live

### User Story

> As an Organization Administrator I would like to add my Org's PluginDefinitions to the Organization's Namespace   

### Description

Each Greenhouse installation has it's own Flux deployment. This will be used to implement the PluginDefinition Catalog. 
The Catalog CRD allows Organization Admins to bring in additional PluginDefinitions into Greenhouse. Also it is used to bring the PluginDefinitions defined in [cloudoperators/greenhouse-extensions](https://github.com/cloudoperators/greenhouse-extensions) to every Organization.

The task is to implement a controller to reconcile a PluginDefinitionCatalog and create Flux resources that will fetch PluginDefinitions from a Github repository and apply them to the Organization's namespace.

The CRD for Catalog should allow different types of sources for the PluginDefinitions. The first type to be implemented is the Github Repository.

### Acceptance Criteria

- ~~[ ] each Org has a ServiceAccount provisioned that provides only CRUD on PluginDefinitions in the org's Namespace [1](https://fluxcd.io/flux/components/kustomize/kustomizations/#role-based-access-control)~~ Replaced by #1321 
- [x] Controller reconciles a Catalog and creates the necessary Flux resources

Issue #1157, opened 2025-05-23
[EPIC] - Reconciliation actions


IvoGoman:
### Description

The reconciliation of Plugin's may need to be triggered manually in certain situations. Currently, the way to go is adding a label/annotation to the resource. So every change to the metadata triggers a reconciliation.
Once the HelmController is replaced with Flux this will not be done so easily, as the Flux resources are only to be changed by the controller.

Actions such as forcing an `helm upgrade`, `helm rollback` or resetting the `retries` on a failed release need to be explicitly triggered.

### Objectives

- Reconciliation of resources (also non-Greenhouse) can be triggered manually
- Special actions required for Plugin releases

### Acceptance Criteria

- [ ] Actions on resources can be triggered via annotations
- [ ] UI can be used to trigger these actions

### Dependencies

- Dependency 1
- Dependency 2
- Dependency 3

PR #1329, opened 2025-08-06
feat(plugindefinitioncatalogs): adds plugin definition catalog controller

abhijith-darshan:
## Description

- Implements `PluginDefintionCatalog`
- Initial resource resolution via Flux `GitRepository` source and `Kustomization` rendering
- Initial support of `github.com` public repositories without authentication

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue #1000 

## Added tests?

- [x] 👍 yes

Unit tests cover -

- Catalog Creation
- Dependent Flux resources creation
- Status Verification

## Added to documentation?

- [ ] 📜 README.md
- [x] (if applicable) generated OpenAPI docs for CRD changes

## Checklist


Makefile
```diff
@@ -350,4 +351,8 @@ cert-manager: kustomize
 
 .PHONY: flux
 flux: kustomize
-	-$(KUSTOMIZE) build config/flux | kubectl apply -f -
+	-$(KUSTOMIZE) build config/samples/flux | kubectl apply -f -
+
+.PHONY: license
+license:
+	docker run --rm -v $(shell pwd):/github/workspace $(IMG_LICENSE_EYE) -c .github/licenserc.yaml header fix
```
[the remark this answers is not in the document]
IvoGoman: Fine to have it separately. My comment was to reuse the new function instead of defining it twice.

internal/controller/catalog/catalog_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	greenhousev1alpha1 "github.com/cloudoperators/greenhouse/api/v1alpha1"
+	"github.com/cloudoperators/greenhouse/internal/flux"
+	"github.com/cloudoperators/greenhouse/internal/lifecycle"
+)
+
+const (
+	gitCatalogSourceSuffix = "-git-repository"
+	kustomizeCatalogSuffix = "-kustomization"
```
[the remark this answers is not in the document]
abhijith-darshan: it is not necessary, as far as I can see there may not be naming collisions.

internal/controller/catalog/catalog_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	if obj.GetDeletionTimestamp().IsZero() {
+		if err = r.Delete(ctx, obj); err != nil {
+			r.recorder.Eventf(catalog, corev1.EventTypeWarning, "Error", "Error: deleting %s for catalog: %s/%s - %s", kind, catalog.GetNamespace(), catalog.GetName(), err.Error())
+			r.log.Error(err, "failed to delete object", "name", obj.GetName(), "namespace", obj.GetNamespace())
+			return
+		}
+	} else {
+		obj.SetAnnotations(map[string]string{
+			greenhouseapis.FluxReconcileRequestAnnotation: time.Now().Format(time.DateTime),
```
uwe-mayer, line 120: We could make this a generic reconcile annotation, not Flux specific?

uwe-mayer: Maybe with a helper function that does the whole time formatting as well?

uwe-mayer (approved):
One non-blocking comment

internal/controller/catalog/catalog_controller.go
IvoGoman: This also falls under https://github.com/cloudoperators/greenhouse/issues/1157 

> Reconciliation of resources (also non-Greenhouse) can be triggered manually

This should read as, we annotate the Plugin and our Flux Controller annotates the underlying Flux resources (if required)

abhijith-darshan: But the thing here is when flux times out and cannot release its resource it won’t reconcile for another 5mins. But if we do this then at least in Unit tests we can easily mock the release and in E2e as well.

but I agree this is a very stressful op on API server. Especially if the CPD / PD is currently in use. Then it may be this way for a very long time. Fortunately in the first release we are not pruning at all. We are going with Prune: false (default) so this shouldn’t matter for now.

I have an idea about spreading the reconcile over based on greenhouse error types. It’s the error standardization we discussed a while back. Where on certain errors it won’t reconcile and spreads evenly.

A similar implementation is in flux as well. It is called as reconcile error stalling. They use it when there is no point in immediate requeue.

https://github.com/fluxcd/source-controller/blob/5f9702bb01267d95c523256bb61115b2d6c68dfd/internal/controller/gitrepository_controller.go#L685

PR #1433, opened 2025-10-01
feat(catalogs): adds github authentication to catalog

### LINKED ISSUE #1322: [FEAT] - Github authentication for Catalog
### Priority

None

### User Story

_No response_

### Description

In order to avoid running into rate-limiting issues when fetching the Git repositories of PluginDefinitionCatalogs it should be possible to configure authentication for Github (Enterprise).

TODO: Evaluate if an org-wide configuration will work for the read-only operations of the Catalog

### Acceptance Criteria

- [x] Authentication can be configured on Catalog
- [x] created GithubRepositories use GithubAuthentication (if provided by Catalog)

### Reference Issues

- https://github.com/fluxcd/pkg/issues/998
- https://github.com/fluxcd/source-controller/issues/1858

abhijith-darshan:
## Description

`Catalog` resource accepts `secretName` in spec to propagate secretReference to flux `GitRepository` for authentication

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #1322 

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no

## Checklist


config/samples/flux/kustomization.yaml
```diff
@@ -17,29 +17,4 @@ patches:
...
-                livenessProbe:
-                  initialDelaySeconds: 30
-                  periodSeconds: 20
-                  timeoutSeconds: 5
-                  failureThreshold: 3
-                readinessProbe:
-                  initialDelaySeconds: 15
-                  periodSeconds: 20
-                  timeoutSeconds: 5
-                  failureThreshold: 3
```
IvoGoman, line 45: This looks unrelated and might slipped through?

internal/controller/catalog/catalog_controller.go
```diff
@@ -318,6 +340,11 @@ func (r *CatalogReconciler) ensureKustomizationIsReady(ctx context.Context, cata
 	kustomization.SetNamespace(catalog.Namespace)
 	err = r.Get(ctx, client.ObjectKeyFromObject(kustomization), kustomization)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			requeue = true
+			err = nil
+			return
+		}
```
IvoGoman, line 347: These checks obfuscate the error. If the controller for some reason is not able to create the Kustomization resource it won't be visible on logs/catalog status etc.
Also what is the difference to just returning the error and let the machinery handle the requeueing?

config/samples/flux/kustomization.yaml
abhijith-darshan: I need flux v2.7 in dev-env. Realized I copied this over from one of the internal repos and it is not needed.

internal/controller/catalog/catalog_controller.go
```diff
@@ -54,10 +60,8 @@ func (r *CatalogReconciler) SetupWithManager(name string, mgr ctrl.Manager) erro
 					predicate.LabelChangedPredicate{},
 				),
 			),
-		).
-		Owns(&sourcev1.GitRepository{}).
-		Owns(&kustomizev1.Kustomization{}).
-		Complete(r)
+		)
+	return r.watchOwnerBuilder(mgr, build, &sourcev1.GitRepository{}, &kustomizev1.Kustomization{}).Complete(r)
```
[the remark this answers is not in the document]
abhijith-darshan: The reason was:

Owns triggers a reconcile on `ensureCreated` lifecycle when flux resources are being deleted. This will cause conflicts as the catalog is undergoing deletion lifecycle and waiting till flux resources are being deleted so we shouldn't attempt a CreateOrPatch at this point. 

We would need to check deletion timestamp on flux resources and ignore reconcile in catalog `ensureCreated`, so in order to avoid Condition flapping of Created | Deleted in lifecycle.Reconcile I moved it to `EnqueueRequestForOwner` 

but I can revert this back to Owns and add a predicate as this works as well - 

```go
func (r *CatalogReconciler) SetupWithManager(name string, mgr ctrl.Manager) error {
	....
		Owns(&sourcev1.GitRepository{}, builder.WithPredicates(ignoreObjectBeingDeleted())).
		Owns(&kustomizev1.Kustomization{}, builder.WithPredicates(ignoreObjectBeingDeleted())).
		Complete(r)
}
func ignoreObjectBeingDeleted() predicate.Predicate {
	return predicate.Funcs{
		UpdateFunc: func(e event.UpdateEvent) bool {
			return e.ObjectNew.GetDeletionTimestamp().IsZero()
		},
		DeleteFunc: func(_ event.DeleteEvent) bool {
			return false
		},
	}
}
```

PR #1530, opened 2025-10-17
feat(catalogs): multi-source catalog

abhijith-darshan:
## Description

Catalog can resolve resources from multiple sources
(Currently targeting multiple GitHub sources)

internally Catalog utilizes, flux `AritfactGenerator` to generate ExternalArtifact, which is consumed by `Kustomization`

## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

- Closes #1525 

## Added tests?

- [x] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help
- [ ] Separate ticket for tests # (issue/pr)

Test to confirm ArtifactGenerator and ExternalArtifact applies `(Cluster)PluginDefinitions`

## Added to documentation?

- [ ] 📜 README.md
- [ ] 🤝 Documentation pages updated
- [ ] 🙅 no documentation needed
- [ ] (if applicable) generated OpenAPI docs for CRD changes

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

Issue #1521, opened 2025-10-14
[FEAT] - Documentation enhancements


IvoGoman:
### Priority

None

### Description

This is a catch-all issue to collect documentation enhancements & gaps.

### Acceptance Criteria

- [ ] Documentation is up-to-date

IvoGoman:
superseded by #1964

PR #1534, opened 2025-10-17
feat(core): add suspend annotation handling

IvoGoman:
## Description

all reconciliation is skipped if suspend annotation is set. This includes regular reconciliation and deletion. Deletion may be handled after the annotation is removed.

For controllers such as Catalog/Plugin the provisioned Flux resources are suspended as well. Ensuring no changes are performed in-cluster or on remote clusters.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -36,7 +36,7 @@ jobs:
       # https://github.com/sigstore/cosign-installer
       - name: Install cosign
         if: github.event_name != 'pull_request'
-        uses: sigstore/cosign-installer@d7543c93d881b35a8faa02e8e3605f69b7a1ce62 # v3.10.0
+        uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
       
       # Set up QEMU for cross-platform builds
       - name: Set up QEMU
@@ -133,7 +133,7 @@ jobs:
           severity: 'CRITICAL,HIGH,MEDIUM'
 
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@42213152a85ae7569bdb6bec7bcd74cd691bfe41 # v3
+        uses: github/codeql-action/upload-sarif@d198d2fabf39a7f36b5ce57ce70d4942944f006e # v3
         if: always()
         with:
           sarif_file: trivy-results.sarif
--- a/charts/cors-proxy/Chart.yaml
+++ b/charts/cors-proxy/Chart.yaml
@@ -18,10 +18,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.2.1
+version: 0.3.0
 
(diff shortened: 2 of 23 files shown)


abhijith-darshan:
Maybe we could do this in lifecycle pkg instead? If suspend annotation exists then each greenhouse object can implement the suspend interface to set the necessary suspend field. Wdyt?

Along the same lines of skip reconcile etc…

IvoGoman:
> Maybe we could do this in lifecycle pkg instead? If suspend annotation exists then each greenhouse object can implement the suspend interface to set the necessary suspend field. Wdyt?
> 
> Along the same lines of skip reconcile etc…

The `EnsureSuspended(..)` is part of the reconciler interface of the lifecycle package. The `Reconcile(..)` ensures suspend logic of the individual controllers is called before anything else. That way the suspending can be handled in `EnsureSuspended(..)` and the `EnsureCreated(..)` ensures that reconciliation is resumed also for child resources.
By default now the Suspend Logic is "just" skipping reconciliation. But for some resources (Plugin/Catalog) it also ensures that child resources will be suspended.
Kept it as an annotation so suspending/resuming the resource would be the same interaction via kubectl. Copying this to a spec Field in the resource would also allow us to interact with the resource from UI/kubectl without a CD tool detecting a drift. 

For the reconcile we could think about keeping the annotation. That way the UI could set a timestamp when a reconcile is requested for the Plugin. So that the controller does the needful. For kubectl interaction the annotation would need to be updated. Then we can in the first place ensure that this annotation will always trigger reconciliation on the resources and we can set the reconcile annotation on managed flux resources if they have changed.

internal/controller/catalog/catalog_controller.go
```diff
@@ -167,6 +168,20 @@ func (r *CatalogReconciler) EnsureCreated(ctx context.Context, obj lifecycle.Run
...
+		return ctrl.Result{}, err
+	}
+
+	if err := r.suspendGitRepository(ctx, catalog); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
```
abhijith-darshan: the catalog controller has significantly changed with multiple sources. And we would need to do it to suspend on per source via spec or all sources via annotations. shall we roll out this functionality after https://github.com/cloudoperators/greenhouse/pull/1530 is merged?

uwe-mayer:
Nice, thx for adding this! 
I am sure you have thought about it: Can we have some kind of docs on how this works?

IvoGoman:
> Nice, thx for adding this! I am sure you have thought about it: Can we have some kind of docs on how this works?

Started to look into refactoring the docs as a whole. New entries are currently logged here: https://github.com/cloudoperators/greenhouse/issues/1521

PR #1571, opened 2025-10-27
feat(controller): add disable-controller flag

kengou:
I’ve noticed that the current `--controllers` flag can be a bit cumbersome when you want to disable a single controller. If you want to disable just one controller, you have to explicitly list all the other controllers in the `--controllers` flag. This can be tedious and error-prone.

To address this, I’ve proposed adding a new flag called `--disabled-controllers`. This flag would allow you to disable one or more controllers without having to list all the other controllers in the `--controllers` flag. This would make the process much more convenient and user-friendly.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


cmd/greenhouse/controllers.go
```diff
@@ -58,16 +58,24 @@ func knownControllersNames() []string {
...
+		if s == str {
 			return true
 		}
 	}
 	return false
 }
 
+// isControllerEnabled checks whether the given controller or regex is enabled
+func isControllerEnabled(controllerName string) bool {
+	if contains(disabledControllers, controllerName) {
```
abhijith-darshan: can be simplified with `slices.Contains(disabledControllers, controllerName)`

cmd/greenhouse/controllers.go
```diff
@@ -58,16 +58,24 @@ func knownControllersNames() []string {
...
+		return false
+	}
+	return controllerName == "*" || contains(enabledControllers, controllerName)
```
abhijith-darshan: same here `slices.Contains(enabledControllers, controllerName)`

cmd/greenhouse/controllers.go
kengou: fixed in [94eb316](https://github.com/cloudoperators/greenhouse/pull/1571/commits/94eb316081e8d33f0cffe0e76e3c70e32e5d2ebb)

kengou:
> > Heyho @kengou ! Thanks for your PR!
> > Can you explain what would happen if both flags are set? Need to have a clear instruction about precedence, right?
> 
> I think that's the idea to simultaneously allow disabling of certain controllers and enabling a sub set of controllers. I guess disabled takes precedence over enabled.

Exactly as @abhijith-darshan explained. Disabled flag has higher priority

PR #1737, opened 2026-01-16
feat(greenhouse): adds dashboards for alerts and plugins

olandr:
This adds two dashboards: Alerts and Plugin. The main design drivers for both of these dashboards has been to (i) use the greenhouse_* metrics, (ii) be a 1-1 mapping of the Greenhouse Alerts, and (iii) give context to the alerts more than simply a number.

The way it has been designed, to not make it completely unusable due to the number of params (e.g. clusterNames, nodes, jobs) is to add UX elements that allows for filtering some params.

---------

Signed-off-by: Simon Olander <simon.olander@sap.com>

## Summary by CodeRabbit

* **New Features**
  * Added three comprehensive monitoring dashboards: "Alerts" dashboard for tracking Kubernetes version updates, resource ownership labels, and cluster readiness; "Organization" dashboard for organization and team status monitoring; "Plugins" dashboard for plugin workload status, reconciliation metrics, and health tracking.


IvoGoman (commented):
Thanks for the big rework @olandr 🎉 
A couple of things which would be cool:

**Organization**
- the Variable Organization does not behave as intended. Can you use  `greenhouse_organization_ready` as the Series selector. (This only being greenhouse will be addressed soon)
- Number of Team Members would be nice as an Timeseries Chart with `{{team}}` in the legend

For the other two let's get them in use them and understand where we can tweak them.

PR #1993, opened 2026-05-18
feat(plugins): load helm chart from flux artifact for oci mirror flow

### LINKED ISSUE #1844: [FEAT] - Use flux helm release to template Plugin manifest for container image replacement
### Priority

(Low) Something is a little off

### User Story

> When running the local environment we want to enable the container image overwriting/mirroring with the local registry

### Description

Currently the Plugin controller pulls a helm release from it's repository to find and replace any mirrored container images. With flux the HelmRelease is already present in-cluster. We do not need to make the external call.

Currently this blocks the dev-env to work with the local registry and image mirroring as the Greenhouse controller does not know the registry's CA.

### Acceptance Criteria

- [ ] Helm release used to template the manifest to filter mirrored container images is pulled from the in-cluster flux HelmRelease resource
- [ ] All tests stay green, this should just be an internal refactoring

### Reference Issues

_No response_

abhijith-darshan:
## Description

This PR replaces the upstream registry pull during plugin reconciliation with a fetch from the Flux source-controller artifact already produced for the plugin's HelmChart. The chart `.tgz` is persisted on disk via a shared `flux.IArtifactory`, then handed directly to the helm chart loader — avoiding repeated upstream pulls during post-render image transformation.

The artifact fetcher (previously catalog-internal) is generalized into a reusable `internal/flux` package. The interface now takes `artifactID` per call, exposes `GetRaw` for raw bytes (no tar/gzip extraction), and adds `Path` so callers can resolve the on-disk location of a cached artifact. Both catalog and plugin reconcilers construct one shared instance in `SetupWithManager`.

The chart now mounts a capped 1Gi emptyDir at the artifact storage path so the on-disk cache is bounded and isolated from the node's writable layer. The CLI flags are renamed from `--catalog-artifact-*` to `--artifact-*` to reflect their broader use, and the storage path is wired from a single values key so the mount and flag cannot drift.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #1844

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md

## Checklist


internal/flux/artifact.go
```diff
@@ -134,13 +164,13 @@ func (a *artifactory) saveToFileSystem(content []byte, digest string) error {
...
 		"digest", digest,
-		"artifactID", a.artifactID,
+		"artifactID", artifactID,
 	)
 	return nil
 }
 
-func (a *artifactory) deleteAllExceptFromFileSystem(keepDigest string) error {
-	dir := filepath.Join(a.storageBasePath, a.artifactID)
+func (a *artifactory) deleteAllExceptFromFileSystem(artifactID, keepDigest string) error {
```
mikolajkucinski, line 172: I think we should have guard against the empty `artifactID`, right? Otherwise we might be operating on the dir root level?

internal/flux/artifact.go
mikolajkucinski: I know its not in the scope of the PR but since we touch these funcs might be worth to change that to `errors.Is(err, os.ErrNotExist)`?

internal/flux/artifact.go
abhijith-darshan: ArtifactID is always available.
`artifactID := helmChart.Namespace + "/" + helmChart.Name`

internal/controller/plugin/plugin_controller.go
```diff
@@ -41,6 +42,9 @@ type PluginReconciler struct {
 	ExpressionEvaluationEnabled bool
 	IntegrationEnabled          bool
 	OCIMirroringEnabled         bool
+	StoragePath                 string
+	HttpRetry                   int
```
IvoGoman: Same case for arconyms
```suggestion
	HTTPRetry                   int
```

cmd/greenhouse/controllers.go
```diff
@@ -88,6 +88,8 @@ func startPluginReconciler(name string, mgr ctrl.Manager) error {
 		ExpressionEvaluationEnabled: featureFlags.IsExpressionEvaluationEnabled(),
 		IntegrationEnabled:          featureFlags.IsIntegrationEnabled(),
 		OCIMirroringEnabled:         featureFlags.IsOCIMirroringEnabled(),
+		StoragePath:                 artifactStoragePath,
+		HttpRetry:                   artifactRetries,
```
IvoGoman: Same case for acronyms
```suggestion
		HTTPRetry:                   artifactRetries,
```

internal/controller/catalog/catalog_controller.go
```diff
@@ -43,12 +44,14 @@ type CatalogReconciler struct {
 	recorder    events.EventRecorder
 	StoragePath string
 	HttpRetry   int
```
IvoGoman: Same case for acronyms:
```suggestion
	HTTPRetry   int
```

IvoGoman (approved):

PR #2014, opened 2026-05-21
feat(plugins): plugin controller deletion lifecycle

Zaggy21:
## Description

Fixes the plugin deletion lifecycle where the greenhouse.sap/cleanup finalizer was removed before Flux had finished running helm uninstall, potentially leaving orphaned Helm resources on the target cluster.

EnsureFluxDeleted now:
- Issues the delete on the HelmRelease and requeues,
- If the HelmRelease is still present, checks Flux's ReleasedCondition: an UninstallFailed reason surfaces as HelmUninstallFailedReason on the Plugin; otherwise sets HelmReleaseUninstallPendingReason and relies on the HelmRelease watch to re-trigger reconciliation,
- Only returns lifecycle.Success (removing the finalizer) once the HelmRelease is gone.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #1885 

## Added tests?

- [x] 👍 yes

Adds unit tests covering the pending and failure paths, and an e2e scenario that injects an uninstall failure via a test finalizer on the HelmRelease and verifies the Plugin surfaces the condition and retains its finalizer until resolved.

## Added to documentation?

- [ ] 📜 README.md

## Checklist


IvoGoman (dismissed):

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,32 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
 	}
 
-	plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmReleaseUninstalledReason, ""))
-	return ctrl.Result{}, lifecycle.Success, nil
+	// Observe the HelmRelease deletion: hold the finalizer until Flux completes the uninstall.
+	hr := &helmv2.HelmRelease{}
+	err := r.Get(ctx, types.NamespacedName{Name: plugin.Name, Namespace: plugin.Namespace}, hr)
+	if apierrors.IsNotFound(err) {
+		// HelmRelease is fully gone — uninstall complete.
+		plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmReleaseUninstalledReason, ""))
```
abhijith-darshan: Why do we need a condition when it is returned success below?

abhijith-darshan (changes requested):

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,32 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
+	released := meta.FindStatusCondition(hr.Status.Conditions, helmv2.ReleasedCondition)
+	if released != nil && released.Reason == helmv2.UninstallFailedReason {
+		msg := released.Message
+		plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmUninstallFailedReason, msg))
+		util.UpdatePluginReconcileTotalMetric(plugin, util.MetricResultError, util.MetricReasonUninstallHelmFailed)
+		return ctrl.Result{}, lifecycle.Failed, fmt.Errorf("helm uninstall failed: %s", msg)
+	}
+
+	// Flux is still running the uninstall; return Pending to keep the finalizer and requeue.
+	plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmReleaseUninstallPendingReason, "waiting for HelmRelease to be removed"))
```
abhijith-darshan: There is a generic deletion condition in place already ...

https://github.com/cloudoperators/greenhouse/blob/2df72613ab6ee66539f862338b1ecddf08cdac8a/pkg/lifecycle/reconcile.go#L213-L228

e2e/plugin/scenarios/flux_controller.go
```diff
@@ -603,3 +603,146 @@ func FluxControllerPluginDeletePolicyRetain(ctx context.Context, adminClient cli
 		g.Expect(err).NotTo(HaveOccurred(), "should be able to uninstall the helm release for the plugin")
 	}).Should(Succeed(), "the retained Helm release should eventually be uninstalled from the remote cluster")
 }
+
+// FluxControllerPluginDeletionLifecycle verifies that the Plugin controller holds its finalizer
+// and surfaces the HelmRelease uninstall failure as a status condition when Flux reports an error.
+func FluxControllerPluginDeletionLifecycle(ctx context.Context, adminClient client.Client, env *shared.TestEnv, remoteClusterName, teamName string) {
```
abhijith-darshan, line 544: Why do we need a dedicated scenario? 😄 

Every scenario should satisfy this

internal/controller/plugin/plugin_controller_flux.go
Zaggy21: Good point. Removed the unnecessary condition.

internal/controller/plugin/plugin_controller_flux.go
Zaggy21: Changed it to use only the generic one.

e2e/plugin/scenarios/flux_controller.go
Zaggy21: It's to cover the HelmRelease uninstall failure scenario as per the acceptance criterion. Can't really add it seamlessly to other scenarios. The basic check for "Plugin stays pending until HelmRelease is deleted" could be added to other scenarios, but it doesn't add much value. It's also covered in unit tests. I'd leave it as it is.

abhijith-darshan: We cannot rely on unit tests. There is no work load in there to guarantee clean ups.

There is already failing plugin e2e scenarios... in one of those we can expect HelmRelease to remain which should already be the case.

A dedicated scenario just for this is not needed.

abhijith-darshan: So to simplify -

During the clean up of every scenario - 

- Expect that the `HelmRelease` is also gone
- in Existing failing scenarios - Expect both `Plugin` and `HelmRelease` remain.

e2e/plugin/e2e_test.go
```diff
@@ -104,6 +104,10 @@ var _ = Describe("Plugin E2E", Ordered, func() {
 		scenarios.FluxControllerPluginDeletePolicyRetain(ctx, adminClient, env, remoteClusterName, team.Name)
 	})
 
+	It("should surface HelmRelease uninstall failure as a Plugin condition and retain the finalizer until resolved", func() {
```
abhijith-darshan, line 107: This is no longer the case? Since the condition is removed

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,30 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
+		return ctrl.Result{}, lifecycle.Success, nil
+	}
+	if err != nil {
+		plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmUninstallFailedReason, err.Error()))
+		util.UpdatePluginReconcileTotalMetric(plugin, util.MetricResultError, util.MetricReasonClusterAccessFailed)
+		return ctrl.Result{}, lifecycle.Failed, err
+	}
+
+	// HelmRelease still exists; check whether Flux reported an explicit uninstall failure.
+	released := meta.FindStatusCondition(hr.Status.Conditions, helmv2.ReleasedCondition)
```
abhijith-darshan:
```suggestion
	releaseStatus := meta.FindStatusCondition(hr.Status.Conditions, helmv2.ReleasedCondition)
```

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,30 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
+	if released != nil && released.Reason == helmv2.UninstallFailedReason {
```
abhijith-darshan:
```suggestion
	if releaseStatus != nil && releaseStatus.Reason == helmv2.UninstallFailedReason {
```

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,30 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
+		msg := released.Message
```
abhijith-darshan:
```suggestion
		msg := releaseStatus.Message
```

abhijith-darshan (changes requested):
minor improvements

e2e/plugin/scenarios/flux_controller.go
Zaggy21: Done. Removed the dedicated scenario and added assertions to FluxControllerPluginDependencies's teardown: after deleting the presets, it verifies both Plugins are marked for deletion before the global plugin is removed and Flux can complete the uninstall.

e2e/plugin/e2e_test.go
Zaggy21: The whole test scenario has been removed.

internal/controller/plugin/plugin_controller_flux.go
```diff
@@ -55,8 +56,30 @@ func (r *PluginReconciler) EnsureFluxDeleted(ctx context.Context, plugin *greenh
...
+	releaseStatus := meta.FindStatusCondition(hr.Status.Conditions, helmv2.ReleasedCondition)
+	if releaseStatus != nil && releaseStatus.Reason == helmv2.UninstallFailedReason {
+		msg := releaseStatus.Message
+		plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousev1alpha1.HelmReleaseDeployedCondition, greenhousev1alpha1.HelmUninstallFailedReason, msg))
+		util.UpdatePluginReconcileTotalMetric(plugin, util.MetricResultError, util.MetricReasonUninstallHelmFailed)
+		return ctrl.Result{}, lifecycle.Failed, fmt.Errorf("helm uninstall failed: %s", msg)
+	}
+
+	// Flux is still running the uninstall; return Pending to keep the finalizer and requeue.
+	return ctrl.Result{RequeueAfter: 10 * time.Second}, lifecycle.Pending, nil
```
abhijith-darshan: The requeue can take a bit longer? _1 minute_...

Since we Own the `HelmRelease`, on clean up we should get the reconcile event?

https://github.com/cloudoperators/greenhouse/blob/6204e4cade5810fcd545e4554f283d8d2d8636fd/internal/controller/plugin/plugin_controller.go#L97

abhijith-darshan: wdyt?

internal/controller/plugin/util.go
```diff
@@ -227,6 +229,11 @@ func shouldReconcileOrRequeue(ctx context.Context, c client.Client, plugin *gree
 	if err != nil {
 		return nil, err
 	}
+	if cluster.GetDeletionTimestamp() != nil {
+		msg := fmt.Sprintf("cluster %s is being deleted", plugin.Spec.ClusterName)
+		plugin.SetCondition(greenhousemetav1alpha1.FalseCondition(greenhousemetav1alpha1.DeleteCondition, lifecycle.PendingDeletionReason, msg))
+		return &reconcileResult{requeueAfter: clusterDeletingRequeueAfter}, nil
+	}
```
abhijith-darshan: so if a Cluster deletion is in progress then the cluster controller should have marked the plugin as deleted so we will not enter the EnsureCreated phase at all right?

Can you reason as to why this was necessary?

abhijith-darshan (changes requested):

internal/controller/plugin/util.go
Zaggy21: I haven't encountered a problem here, I just tend to be overeager with defensive guards :D But I can see now that it's an overkill, the `shouldReconcileOrRequeue` is actually on the top of the reconciliation loop and race condition is not likely. I'll remove it.

internal/controller/plugin/plugin_controller_flux.go
Zaggy21: Yes, the`Owns(&helmv2.HelmRelease{})` should be enough. I'm convinced to remove the `requeueAfter` here entirely, because we will get an event that will trigger the reconciliation when the helm release is uninstalled.

abhijith-darshan (approved):

Issue #1776, opened 2026-02-03
[FEAT] - PluginPreset controller evaluates PluginOptionValue.Ref


IvoGoman:
### Priority

None

### Description

PluginPreset controller takes over the evaluation of references to other Plugins. Plugin controller will no longer evaluate references.
The Plugins created from the PluginPreset controller have no dynamic PluginOptionValues.

### Acceptance Criteria

- [ ] Plugin created from a PluginPreset have their PluginOptionValue references resolved
- [ ] Logic for evaluation of PluginOptionValue references moves from Plugin to PluginPreset controller

PR #2117, opened 2026-07-03
feat(pluginpresets): pluginpreset controller resolves values from references

k-fabryczny:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/cloudoperators/greenhouse/issues/1776

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [x] 🤝 Documentation pages updated

## Checklist


internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -5,37 +5,55 @@ package plugin
...
 	if r.ExpressionEvaluationEnabled {
-		return r.resolveExpressionsForPreset(ctx, preset, cluster)
+		var err error
+		resolvedValues, err = r.resolveExpressionsForPreset(ctx, preset, cluster)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve expressions: %w", err)
+		}
+	} else {
+		resolvedValues = util.ConvertToPluginOptionValues(preset.Spec.Plugin.OptionValues)
 	}
```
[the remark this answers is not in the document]
k-fabryczny: The Expression and ValueFrom.Ref fields have been removed from PluginOptionValue and PluginValueFromSource. The current implementation correctly copies all existing fields.

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -5,37 +5,55 @@ package plugin
...
+	if r.IntegrationEnabled {
+		var err error
+		resolvedValues, err = r.resolveReferencesForPreset(ctx, cluster, preset.Namespace, preset.Spec.Plugin.OptionValues, resolvedValues)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve references: %w", err)
 		}
 	}
 
-	return util.ConvertToPluginOptionValues(preset.Spec.Plugin.OptionValues), nil
+	return resolvedValues, nil
```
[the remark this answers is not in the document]
k-fabryczny: Check for expression exists: 
`		for _, ov := range preset.Spec.Plugin.OptionValues {
			if ov.Expression != nil {
				return nil, fmt.Errorf("option %s has expression but expressionEvaluationEnabled is disabled for PluginPreset controller", ov.Name)
			}`

docs/reference/components/pluginpreset.md
```diff
@@ -196,6 +197,173 @@ data:
...
+The expression field in valueFrom.ref supports multiple syntax styles:
+
+#### New simplified syntax
+```expression: spec.optionValues.filter(v, v.name == "my.value")[0].value```
+
+#### With ${...} wrapper
+```expression: ${spec.optionValues.filter(v, v.name == "my.value")[0].value}```
+
+#### Legacy syntax (backward compatible)
+```expression: object.spec.optionValues.filter(v, v.name == "my.value")[0].value```
```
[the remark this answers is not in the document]
k-fabryczny: Done

docs/reference/components/pluginpreset.md
```diff
@@ -176,13 +174,16 @@ spec:
           "special.${global.greenhouse.metadata.region}.example.com"
 ```
 
-> :information_source: Expressions are evaluated in PluginPresets when `pluginPreset.expressionEvaluationEnabled` is enabled.
-Standalone Plugin expressions are still supported (deprecated) and may be evaluated by the Plugin controller depending on feature flags.
+> :information_source: Expressions are only evaluated in PluginPresets.
+
+> :warning: CEL expressions on standalone Plugins are deprecated and will be removed in a future release. Use PluginPresets for expression evaluation.
```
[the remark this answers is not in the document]
k-fabryczny: Done

charts/manager/templates/_helpers.tpl
```diff
@@ -113,4 +113,8 @@ Define postgresql helpers
 {{/* Render the pluginPreset expression evaluation flag */}}
 {{- define "pluginPreset.expressionEvaluationEnabled" -}}
   {{- printf "%t" (required "global.pluginPreset.expressionEvaluationEnabled missing" .Values.global.pluginPreset.expressionEvaluationEnabled) }}
+{{- end }}
+{{/* Render the pluginPreset integration enabled flag */}}
+{{- define "pluginPreset.integrationEnabled" -}}
+  {{- printf "%t" (required "global.pluginPreset.integrationEnabled missing" .Values.global.pluginPreset.integrationEnabled) }}
```
mikolajkucinski, line 119: i think this breaks `helm template` for the manager chart, charts/manager/ci/test-values.yaml doesn't have this key

internal/controller/plugin/pluginpreset_controller.go
```diff
@@ -48,6 +48,7 @@ type PluginPresetReconciler struct {
 	client.Client
 	recorder                    events.EventRecorder
 	ExpressionEvaluationEnabled bool
+	IntegrationEnabled          bool
```
mikolajkucinski, line 52: is there anything that re-reconciles the consumer when the referenced preset changes, i can't find it

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,275 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+	if err := r.List(ctx, presetList,
+		client.InNamespace(namespace),
+		client.MatchingLabelsSelector{Selector: selector},
+	); err != nil {
+		return nil, fmt.Errorf("failed to list PluginPresets by selector: %w", err)
+	}
+
+	if len(presetList.Items) == 0 {
+		log.Info("No PluginPresets found matching selector", "selector", ref.Selector)
+		return []any{}, nil
```
mikolajkucinski: if the source presets aren't created yet this quietly deploys an empty list no?

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,275 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+
+// resolveReferencedPresetValues resolves expressions in a referenced PluginPreset
+// if the ExpressionEvaluationEnabled flag is set.
+func (r *PluginPresetReconciler) resolveReferencedPresetValues(
+	ctx context.Context,
+	refPreset *greenhousev1alpha1.PluginPreset,
+	cluster *greenhousev1alpha1.Cluster,
+) []greenhousev1alpha1.PluginOptionValue {
+
+	if !r.ExpressionEvaluationEnabled {
```
mikolajkucinski, line 473: overrides get skipped when only the expression flag is off, is that on purpose?

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,275 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+		return util.ConvertToPluginOptionValues(refPreset.Spec.Plugin.OptionValues)
+	}
+
+	// Apply cluster-specific overrides to referenced preset first
+	refPresetWithOverrides := applyOverridesToPreset(refPreset, cluster.Name)
+
+	resolvedRefValues, err := r.resolveExpressionsForPreset(ctx, refPresetWithOverrides, cluster)
+	if err != nil {
+		log := ctrl.LoggerFrom(ctx)
+		log.Error(err, "Failed to resolve expressions in referenced PluginPreset, using raw values",
```
mikolajkucinski: why fall back to raw values here instead of just returning the error?

mikolajkucinski (commented):
had a read through, few questions below, mostly trying to understand the intent rather than blocking anything

charts/manager/templates/_helpers.tpl
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_controller.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

mikolajkucinski (dismissed):
thanks for addressing my questions/comments :) 
at this stage lgtm, might be good to also get Abhi's/Uwe's approval 🚀

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+	refKind := ref.Kind
+	if refKind == "" {
+		refKind = greenhousev1alpha1.PluginPresetKind
+	}
+
+	switch refKind {
+	case greenhousev1alpha1.PluginPresetKind:
+		return r.resolvePluginPresetRef(ctx, ref, cluster, namespace)
+	default:
+		return nil, fmt.Errorf("unsupported reference kind: %s", refKind)
```
abhijith-darshan, line 219: why the restriction of Plugin? it is contradicting the API -

https://github.com/cloudoperators/greenhouse/blob/217e9669d1c6e08288b84b3b8f91a0ce423b5f3c/api/v1alpha1/plugin_types.go#L80-L101

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+			})
+		} else {
+			result = append(result, resolvedValues[i])
+		}
+	}
+
+	return result, nil
+}
+
+// resolveRef resolves a reference to another resource (PluginPreset or Plugin).
```
abhijith-darshan, line 200: seems like this func explicitly denies Plugin kind but the comment says otherwise

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+			result = append(result, greenhousev1alpha1.PluginOptionValue{
+				Name:  presetOV.Name,
+				Value: &apiextensionsv1.JSON{Raw: byteVal},
+			})
+		} else {
+			result = append(result, resolvedValues[i])
+		}
+	}
+
+	return result, nil
```
abhijith-darshan, line 197: result correlates to two independently-built slices by positional index (resolvedValues[i] vs presetOptionValues[i]). This holds only because neither slice is sorted and both iterate the same source in the same order — this enforcement is not obvious. 

Key by Name instead (it's already present on every entry), so this stays correct if resolution order ever changes or the resolved slice gets sorted.

```go
  // index resolved values by Name before the loop
  resolvedByName := map[string]PluginOptionValue{}
  for _, rv := range resolvedValues {
      resolvedByName[rv.Name] = rv
  }

  for _, presetOV := range presetOptionValues {   // drop index i
      if presetOV.ValueFrom != nil && presetOV.ValueFrom.Ref != nil {
          // ... unchanged ...
      } else {
          rv, ok := resolvedByName[presetOV.Name]  // instead of resolvedValues[i]
          if !ok {
              return nil, fmt.Errorf("resolved value for option %s not found", presetOV.Name)
          }
          result = append(result, rv)
      }
  }
```

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal resolved value for %s: %w", presetOV.Name, err)
+			}
+
+			result = append(result, greenhousev1alpha1.PluginOptionValue{
+				Name:  presetOV.Name,
+				Value: &apiextensionsv1.JSON{Raw: byteVal},
+			})
+		} else {
+			result = append(result, resolvedValues[i])
```
abhijith-darshan: NIT: final result should be sorted otherwise `controllerutil.CreateOrUpdate` does a Get, runs mutate fn, if what's applied doesn't match what's computed then there will be a const reconcile and update.

abhijith-darshan: for now since OptionValues are an array it is fine since arrays preserve order. But nevertheless sorting would be a defensive measure?

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+	for _, ov := range optionValues {
+		item := map[string]any{
+			"name": ov.Name,
+		}
+		if ov.Value != nil && len(ov.Value.Raw) > 0 {
+			var val any
+			if err := json.Unmarshal(ov.Value.Raw, &val); err == nil {
+				item["value"] = val
+			}
+		}
```
abhijith-darshan, line 498: errors are silently ignored here dropping values without logs, warnings on conditions.... this would be hard to debug. We should hard error and not proceed further as incomplete values for a helm chart would cause more issues.

internal/controller/plugin/pluginpreset_values_resolver.go
```diff
@@ -98,9 +131,277 @@ func (r *PluginPresetReconciler) resolveExpressionsForPreset(
...
+		if err != nil {
+			return nil, err
+		}
+		celObject := buildCELObject(refPreset.Name, refPreset.Namespace, resolvedRefValues)
+
+		value, err := evaluateCELWithObject(ref.Expression, celObject)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate reference expression for PluginPreset %s: %w", refPreset.Name, err)
+		}
+		results = appendToResults(results, value)
```
abhijith-darshan, line 456: here ordering / sorting matters - `client.List` doesn't return sorted results, it returns randmon order that means the ordering of result (values resolved from all matching presets) will change randomly causing unnecessary reconcile.

abhijith-darshan: perhaps something like this in `appendToResults` - 

```go
  slices.SortFunc(results, func(a, b any) int {
      ba, _ := json.Marshal(a)
      bb, _ := json.Marshal(b)
      return bytes.Compare(ba, bb)
  })
```

internal/controller/plugin/pluginpreset_controller.go
```diff
@@ -485,6 +491,16 @@ func (r *PluginPresetReconciler) enqueuePluginPresetsForClusterPluginDefinition(
...
+// enqueueReferencingPluginPresets enqueues all PluginPresets in the same namespace
+// when a PluginPreset changes, so that consumers with valueFrom.ref references
+// pick up changes in referenced presets.
+func (r *PluginPresetReconciler) enqueueReferencingPluginPresets(ctx context.Context, obj client.Object) []ctrl.Request {
+	if !r.IntegrationEnabled {
+		return nil
+	}
+	return listPluginPresetAsReconcileRequests(ctx, r.Client, client.InNamespace(obj.GetNamespace()))
+}
+
```
abhijith-darshan: this is very costly when there are 1000s of objects. we need to find a better way.

internal/controller/plugin/pluginpreset_controller.go
abhijith-darshan: errors returned from `r.resolvePluginOptionValuesForPreset(...)` break the entire preset. 

this contradicts err check in line 266 where we surface up the error. 

we need to do the same for `resolvePluginOptionValuesForPreset` errors

abhijith-darshan (commented):
there are quite a bit of short-comings from original implementation.

- Plugin kind is not supported
- Restrictions on CEL extraction only on optionValues (no clusterOptionOverrides, etc...)
- `.status` fields are not considered as an example we would need that for exposed services when there is no ingress available
- affected resource tracking

I would comment on feature-toggle that it is in early stages and experimental and doesn't satisfy all scenarios.

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_controller.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_controller.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/controller/plugin/pluginpreset_values_resolver.go
k-fabryczny: Done

internal/webhook/v1alpha1/pluginpreset_webhook.go
```diff
@@ -137,7 +137,7 @@ func convertPresetToPluginOptionValues(presetValues []greenhousev1alpha1.PluginP
 			Name:  pv.Name,
 			Value: pv.Value,
 		}
-		if pv.ValueFrom != nil {
+		if pv.ValueFrom != nil && pv.ValueFrom.Secret != nil {
```
abhijith-darshan, line 140: so this is to prevent value resolution from cel for secrets?

abhijith-darshan (approved):

PR #2217, opened 2026-09-16
fix(service-proxy): switch to use lifecycle package

jellonek:
Use lifecycle package instead of clientutil for fetching remote kubeconfig.


abhijith-darshan:
superseded by #2220

